PR #5, opened 2019-08-15
Remove web.py for testing migration

joshmoore:
This plugin has been filter branched with
https://github.com/ome/omero-web


joshmoore:
web.py has now been merged to omero-web. Removing here.

PR #6, opened 2019-08-28
5.5.dev2 with Ice deployment 

joshmoore:
In preparation for final push of 5.5.1 to PyPI. Should be merged and released _before_ https://github.com/ome/omero-web/pull/7. Currently installs no dependencies.

cc: @ome/python


setup.py
```diff
@@ -68,23 +44,45 @@
...
-    download_url=url,
     package_dir={"": "target/"},
     packages=packages,
     package_data={
         'omero.gateway': ['pilfonts/*'],
         'omero.gateway.scripts': ['imgs/*']},
-    cmdclass={'test': PyTest},
     scripts=glob.glob(os.path.sep.join(["bin", "*"])),
+    install_requires=[
+        # No Ice included, since a native wheel is preferable
```
manics: I thought if the package/wheel is already installed this will detect it?

joshmoore: If I understand, you're saying that listing Ice here would prove that it had been installed? If so, interesting and happy to discuss. The downside is that then we'll need to figure out how to make travis happily install Ice which I've so far failed to do.

manics: That's my understanding, can you link me to a failing travis build?

manics:
Apart from the zeroc-ice dependency generally seems OK:
```
docker run -it --rm ubuntu:18.04

apt-get update
apt-get install -y python virtualenv

virtualenv /omero-py

/omero-py/bin/pip install https://github.com/ome/zeroc-ice-ubuntu1804/releases/download/0.1.0/zeroc_ice-3.6.4-cp27-cp27mu-linux_x86_64.whl
/omero-py/bin/python -c 'import Ice; print Ice.stringVersion()'
/omero-py/bin/pip install https://github.com/ome/omero-py/archive/ee340a8f85b012ba22efd91a07bcc79b8984229d.zip

useradd -m test
su - test
. /omero-py/bin/activate

omero login ...
```
Note https://github.com/ome/omero-py/issues/9

setup.py
manics: Testing locally with `zeroc-ice>=3.6.4,<3.7`, `zeroc-ice` is picked up as expected:
```
sed -i.bak "s/# No Ice included.*/'zeroc-ice>=3.6.4,<3.7'/" setup.py

pip install .

DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Processing /omero-py-ee340a8f85b012ba22efd91a07bcc79b8984229d
Requirement already satisfied: zeroc-ice<3.7,>=3.6.4 in /omero-py/lib/python2.7/site-packages (from omero-py==5.5.1) (3.6.4)
Building wheels for collected packages: omero-py
  Building wheel for omero-py (setup.py) ... done
...
```

joshmoore: > That's my understanding, can you link me to a failing travis build?

Unfortunately, I've only seen it happen at _deploy_ time, making it difficult to test: https://travis-ci.org/ome/omero-web/builds/577458811#L1909 An alternative option is that we simply don't use travis to deploy.

> zeroc-ice is picked up as expected:

So you're proposing having `zeroc-ice>=3.6.4,<3.7` in all of our setup.py files?

manics: For ease of maintenance I'd only add it to the modules that have a direct dependency (probably omero-py only?), since in principle omero-web 6 could be compatible with omero-py 5.5, and the Ice version is determined by omero-py

manics:
See
- https://github.com/joshmoore/omero-py/pull/2
- https://github.com/joshmoore/omero-py/pull/3

PR #7, opened 2019-08-28
Split gateway

joshmoore:
While running `futurize` &c. on `omero/gateway/__init__.py` unicode errors were being thrown. In order to help identify the cause, this tackles the long-outstanding issue of line count in that file by splitting the gateway into multiple modules.

At the moment, the design is such that all of the new modules get re-imported into omero.gateway so that an end user would not notice the difference. Eventually we should deprecated that usage.


joshmoore:
--exclude

Still too much low-level magic. Excluding.

joshmoore:
Closing in favor of gh-20.

PR #15, opened 2019-09-02
Remove package WARNING

joshmoore:
```
WARNING: '' not a valid package name
```

was printed on every call to `setup.py` meaning that
`python setup.py --version` could not be used for
build-infra integration.


joshmoore:
Merging as simple fixes. Note: this change was also tested as in https://github.com/ome/omero-web/pull/9#issue-313147872 to show that no files went missing.

PR #16, opened 2019-09-02
Revert to manylinux to speed up builds

joshmoore:
see:
 - https://github.com/ome/zeroc-ice-py-manylinux#using-manylinux-packages
 - https://github.com/ome/zeroc-ice-py-manylinux/blob/master/test/Dockerfile.centos7#L7-L8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ services:
   - docker
 
 before_install:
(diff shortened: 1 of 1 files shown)


joshmoore:
Ooh. Didn't realize this was still open. Merging to speed up the builds.

PR #18, opened 2019-09-05
Allow OMERO.py to connect to OMERO websockets with chained SSL certs

manics:
`curl -sSL https://github.com/ome/openmicroscopy/pull/6098.patch | sed 's%/components/tools/OmeroPy/%/%g' | git am`

Ported from https://github.com/openmicroscopy/openmicroscopy/pull/6098


joshmoore:
Seems straight-forward to get in. Merging.

PR #19, opened 2019-09-06
Py3 (1)

joshmoore:
Initial infrastructure for running tests on omero-py with both python2 and python3. Primary requirement is to have both versions installed on a single system. The docker does this for centos7 by using:

```
RUN yum install -y centos-release-scl \
 && yum install -y rh-python36 \
```

which places a Python 3 interpreter under `/opt/rh/rh-python36/root/bin/`.  Once both interpreters are available on a given system, these are the steps to follow:

 * Create a virtualenv for both platforms, e.g.:

```
virtualenv /py2 && /py2/bin/pip install -U pip tox future wheel
export PATH=/opt/rh/rh-python36/root/bin/:$PATH
python -m venv /py3 && /py3/bin/pip install -U pip tox future wheel
```

 * Upgrade pip and install tox in the virtualenvs, e.g.:

```
/py2/bin/pip install -U pip tox
/py3/bin/pip install -U pip tox
``` 
 
 * Run tox which performs the rest of the installations in a fresh virtualenv every time:

```
/py3/bin/tox                  # Creates two environments and runs them both
/py3/bin/tox -e py36          # Tests just the python 3 environment
/py3/bin/tox -- -svxk pattern # Replaces the `posargs` section of tox.ini
```

If you would like to omit `tox` from the process above, then additionally install pytest and the other dependencies from tox.ini into each of the virtualenvs above and use pytest directly.


src/omero/rtypes.py
```diff
@@ -296,15 +296,15 @@ def rstring(val):
...
     elif isinstance(val, omero.RString):
         return val
-    elif isinstance(val, StringTypes):
+    elif isinstance(val, basestring):
         if len(val) == 0:
             return remptystr
         else:
-            if isinstance(val, unicode):
+            if isinstance(val, str):
                 val = val.encode("utf-8")
```
will-moore: Do we still need to do this encoding?

will-moore:
```
$ docker build .
...
...
[various syntax errors] e.g...
  File "/src/.tox/py36/lib/python3.6/site-packages/omero_py-5.5.dev3-py3.6.egg/omero/util/upgrade_check.py", line 142
    except Exception, e:
                    ^
SyntaxError: invalid syntax

  File "/src/.tox/py36/lib/python3.6/site-packages/omero_py-5.5.dev3-py3.6.egg/omero_ext/mox.py", line 779
    raise NotImplementedError, 'method must be implemented by a subclass.'
                             ^
SyntaxError: invalid syntax
...
...
target/omero/rtypes.py:30: in <module>
    IceImport.load("omero_Scripts_ice")
target/IceImport.py:22: in load
    __import__(target)
E     File "/src/target/omero_Scripts_ice.py", line 454
E       """
E        ^
E   SyntaxError: invalid escape sequence \[
!!!!!!!!!!!! xdist.dsession.Interrupted: stopping after 1 failures !!!!!!!!!!!!!
=============================== 1 error in 1.71s ===============================
ERROR: InvocationError for command /src/.tox/py36/bin/pytest -x -n1 -m 'not broken' -rf test/unit/clitest/ test/unit/fstest/ test/unit/gatewaytest/ test/unit/scriptstest/ (exited with code 2)
___________________________________ summary ____________________________________
ERROR:   py27: commands failed
ERROR:   py36: commands failed
FIXME: temporarily passing
Removing intermediate container 99791bcad9f3
 ---> d634da9ca792
Successfully built d634da9ca792
```

manics:
`mox` is incompatible with Python 3, try https://pypi.org/project/mox3/ instead (though it may have bugs, e.g. https://travis-ci.org/ome/omego/jobs/581298560).

joshmoore:
> mox is incompatible with Python 3

Thanks, that's good to know, but aren't the two errors printed valid, i.e. they need futurizing?

> E     File "/src/target/omero_Scripts_ice.py", line 454
E       """
E        ^

My understanding was that this was more of an Ice issue.  https://github.com/ome/omero-blitz/pull/68 was my attempt to fix it. I'll get that into a build as soon as possible.

manics:
Yes, all three problems need fixing

joshmoore:
Tests have now reached:

```
>       if id is not None and isinstance(id, (str, unicode)) and ":" in id:
E       NameError: name 'unicode' is not defined
```

which requires https://github.com/ome/omero-blitz/pull/69 Now need a fix for pointing to 5.5.4-SNAPSHOTs (e.g.) _a la_ https://github.com/ome/omero-py/pull/21

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#55](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/55/). See the [console output](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/55/consoleText) for more details.
Possible conflicts:
  - PR #16 joshmoore 'Revert to manylinux to speed up builds'
    - .travis.yml

--conflicts

joshmoore:
Now green with xfails for Python 2 only (see gh-25)

The changes here will need to be reviewed before this is released, but in order to have a single branch to work from, merging this as the basis for moving to `5.6.dev1` as discussed this morning.

src/omero/rtypes.py
```diff
@@ -296,15 +301,16 @@ def rstring(val):
...
         else:
-            if isinstance(val, unicode):
-                val = val.encode("utf-8")
+            if sys.version_info < (3, 0, 0):
+                if isinstance(val, str):
+                    val = val.encode("utf-8")
             return RStringI(val)
     else:
-        return rstring(unicode(val))
+        return rstring(str(val))
```
will-moore: This is now failing for me:
```
>>> password = "ążćę"
>>> from omero.rtypes import rstring
>>> rstring(password)
  File "/Users/willadmin/Desktop/PYTHON/omero-py/src/omero/rtypes.py", line 310, in rstring
    val = val.encode("utf-8")
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 0: ordinal not in range(128)
```
in integration tests: ```./setup.py test -t test/integration/clitest/test_user.py -k testPassword```.

manics:
Websockets are broken on this branch:
```
$ omero login -s ws://idr.openmicroscopy.org/omero-ws
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
Username: 
Password:
-! 10/07/19 09:23:02.266 warning: Proxy contains unknown endpoints: `//idr.openmicroscopy.org/omero-ws'
WARNING:omero.client:..Ignoring error in client.__del__:<class 'Ice.DNSException'>
Ice.DNSException: bad host name: 'ws://idr.openmicroscopy.org/omero-ws'
```

PR #24, opened 2019-09-16
Bump to blitz python zip 5.5.5

joshmoore:
A few changes to combined.vm allow model
objects to be created which work on Py2
and Py3.


joshmoore:
Proactively merging in order to deal with the inevitable conflicts in https://github.com/ome/omero-py/pull/19

PR #27, opened 2019-09-27
host basestring fix

will-moore:
With other branches, host is type <class 'future.types.newstr.newstr'> and the
check that isinstance(host, str) fails.
This leads to host being set to <'omero.host' not set> which results in the
Ice.EndpointParseException being thrown.

Confusingly, the ```omero-web @login_required``` decorator logs but otherwise ignores the
exception thrown by ```conn = ctx.get_connection(server_id, request)``` and passes ```conn=None``` to the wrapped method, so that the error the user sees is ```None has no attribute getEventContext```.

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -28,6 +28,8 @@ import re
 import ssl
 import uuid
 
+from past.builtins import basestring
+
 IceImport.load("Glacier2_Router_ice")
 import Glacier2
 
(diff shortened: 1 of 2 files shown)


will-moore:
@joshmoore I added 70aeece to make travis happy, but it will conflict with other PRs where it's already added.

will-moore:
Opening against Josh's py3_1 branch instead.

PR #29, opened 2019-09-27
ServiceOpts setitem() ensures value is a string

will-moore:
This fixes
```File "/Users/willadmin/Desktop/py2/omeroweb/lib/python2.7/site-packages/omero_api_IQuery_ice.py", line 651, in findByQuery
    return _M_omero.api.IQuery._op_findByQuery.invoke(self, ((query, params), _ctx))
ValueError: context value must be a string``` in the current merge build with python 2, where we have ```from builtins import str```.

NB: this will likely conflict with the branch that adds that.

### DIFF
--- a/src/omero/gateway/utils.py
+++ b/src/omero/gateway/utils.py
@@ -25,6 +25,7 @@
 
 import logging
 import json
+from future.utils import native_str
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


will-moore:
Closing, to open against @joshmoore py3 branch instead...

Issue #30, opened 2019-10-08
Websockets are broken on this branch due to py3 work


manics:
Websockets are currently broken due to https://github.com/ome/omero-py/pull/19
```
$ omero login -s ws://idr.openmicroscopy.org/omero-ws
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
Username: [root]public
Password:
-! 10/08/19 15:55:56.889 warning: Proxy contains unknown endpoints: `//idr.openmicroscopy.org/omero-ws'
WARNING:omero.client:..Ignoring error in client.__del__:<class 'Ice.DNSException'>
Ice.DNSException: bad host name: 'ws://idr.openmicroscopy.org/omero-ws'
```

PR #32, opened 2019-10-09
hostmatch url regex: don't double escape

manics:
Closes https://github.com/ome/omero-py/issues/30

Testing: `omero login ws://idr.openmicroscopy.org/omero-ws` should create a session (note that output from `omero login` is broken in the current branch as can be seen from a non-websocket login without this PR


manics:
Included in https://github.com/ome/omero-py/pull/39

PR #35, opened 2019-10-10
conn.chgrpObjects() tries to create/link new tags to match old ones

will-moore:
See https://forum.image.sc/t/lost-tags-when-moving-projects-between-groups/30228/2

This attempts to identify tags on Objects being moved to a new Group, look for Tags of a similar name that are in the target group and link them as part of the chgrp process.
If a matching Tag is not found in the target group, an new one is created.

NB: Doesn't check whether the found tag can be linked (e.g. if the target group is read-only).
NB: This only works with the objects selected to be moved, e.g Image, Dataset etc and NOT on any children in the Dataset.
NB: If the Tag is ONLY on the object being moved (and is not linked to anything else in the original group) then the Tag will also change groups and you will have duplicate Tags in the target group.

This should probably be an option on the conn.chgrpObjects() method (instead of happening by default) and that option should also be exposed in the webclient UI.

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -4513,6 +4513,37 @@ class _BlitzGateway (object):
                 save.obj = link
                 saves.append(save)
 
+        # If we are moving taggable objects...
+        if getattr(omero.model, "%sAnnotationLinkI" % graph_spec):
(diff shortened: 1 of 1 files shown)


jburel:
Closing this PR, it prevents other PR attempting to fix the tests to get in

PR #36, opened 2019-10-15
Test natively in travis instead of docker

manics:
The docker tox travis check has the problem that the output of both environments is in the same log so it's not easy to see what's broken. This switches to using Travis environments instead.
- quietens the `python setup.py install` step inside tox to reduce the log output and make it easier to view the test output
- speeds up the time to run the tests in travis which has recently increased due to previously disabled unit tests being enabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,24 +2,34 @@ sudo: required
 
 language: python
 
-python:
-  - "2.7"
+matrix:
+  include:
+    - python: "2.7"
+      env: TOX=py27
+    - python: "3.6"
+      env: TOX=py36
+    - python: "3.6"
+      env: TOX=no
 
-services:
-  - docker
-
-before_install:
-  - pip install -U pip future
-  - pip install https://github.com/ome/zeroc-ice-py-manylinux/releases/download/0.0.10/zeroc_ice-3.6.4-cp27-cp27mu-manylinux2010_x86_64.whl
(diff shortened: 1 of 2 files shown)


joshmoore:
A benefit of having everything defined once in the docker is that for anyone who uses it, checks like rst linting are done before opening a PR. You ok for me to add a commit to have two separate jobs invoking docker separately for each Python version?

manics:
Yes, alternatively have the Docker image prebuilt somewhere, and have Travis run something like `docker run --rm -v $PWD/src:/src omero-py-test-docker tox -e pyXX` for each matrix build.

joshmoore:
re: https://github.com/travis-ci/travis-ci/issues/5358#issuecomment-542672812

> Docker caching can be simplified to just:
```
cache:
  directories:
  - docker_images

before_install:
- docker load -i docker_images/images.tar || true

before_cache:
- docker save -o docker_images/images.tar $(docker images -a --filter="dangling=false" -q)
```

Might be worth a try. I'll take a look.

joshmoore:
@manics : I tested that this worked in docker without the `basepython` line, likely due to the setting of PATH. I'm happy for this to go in as is. (The docker simplification requires pushing to s3 or similar which doesn't seem like it at this point)

Anything else from your side?

manics:
Nothing else. When things are stable we can probably get away with fixed image pulled from Docker Hub.

PR #37, opened 2019-10-16
tables: fix table creation (string issues)

joshmoore:
The native HDF5 libraries are like the Ice libraries
and expect traditional objects (bytes not newbytes).
With this PR, it's possible run the integration tests.
It does *not* fix the issues with ints and longs.


joshmoore:
WFM. Merging. And restarting travis on all failing jobs.

PR #38, opened 2019-10-16
scripts: fix script launching (string issues)

joshmoore:
Fixes Ice context issues as well as the __import__
method, which is another function which solely
takes native strings.


joshmoore:
Merging proactively since integration tests are now running.

PR #34, opened 2019-10-10
More unit tests

manics:
More unit test fixes
Test `test/unit/*.py` (previously ignored in tox.ini)

Amongst other things this should fix some of the problems with displaying tabular CLI output

PR #39, opened 2019-10-16
More unit tests

manics:
Also allows `OMERODIR` to override the path traversal search by testlib for `bin/omero`
https://github.com/ome/omero-py/commit/783f7a38dca95aade4f2e3e032d8c0c6968f45a6 should resolve https://github.com/ome/omero-py/pull/34#issuecomment-542220641


src/omero/testlib/__init__.py
```diff
@@ -188,7 +188,7 @@ def travers(p):
         searched = []
         p = path(".").abspath()
         # "" means top of directory
-        dist_dir = None
+        dist_dir = os.getenv('OMERODIR')
```
joshmoore: Nice idea.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#84](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/84/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/84/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - tox.ini

--conflicts

src/omero/testlib/__init__.py
joshmoore: ...but we need a `path` rather than a `str`:

```
    def add_client_dir(self):
>       client_dir = self.omero_dist / "lib" / "client"
E       TypeError: unsupported operand type(s) for /: 'str' and 'str'
```

manics: That's done in the line below. Where are you seeing this error? The only `add_client_dir` method in this repo is different: https://github.com/ome/omero-py/blob/6199457873bc501578ed3f60f78c920ddaf93587/test/unit/clitest/test_import.py#L61

joshmoore: My understanding is that since `dist_dir` is not `None` with the new logic, that a `str` is getting returned and then used as `self.omero_dist`.

sbesson: It looks like the error lineis https://github.com/ome/openmicroscopy/blob/d0ba94dac3b50ed4bc379dac334b4d5e92445ae9/components/tools/OmeroPy/test/integration/clitest/test_import.py#L157 i.e. `ImportTest` subclasses `CLITest` and overrides `add_client_dir`

joshmoore:
Merging proactively so we can avoid conflicts.

PR #43, opened 2019-10-17
sessions

jburel:
* fix issue when running ``bin/omero sessions login``
* Tested on both Python 2 and 3

Tested the various option. I only found  a problem with ``login``

### DIFF
--- a/src/omero/plugins/sessions.py
+++ b/src/omero/plugins/sessions.py
@@ -447,14 +447,14 @@ class SessionsControl(UserGroupControl):
                                      " %s:%s" % (previous[1], previous[0],
                                                  previous[3]))
                         self.ctx.dbg("Exception on attach: %s"
-                                     % traceback.print_exc(e))
+                                     % traceback.format_exception(None, e, e.__traceback__))
                         try:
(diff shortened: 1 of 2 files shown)


joshmoore:
:+1: good catch. I'll add a fix for the one other example to this PR:

```
/opt/omero-py $git grep -E 'format_exc[(]\w+[)]'
src/omero/plugins/sessions.py:                                     % traceback.format_exc(e))
src/omero/plugins/sessions.py:                                         % traceback.format_exc(e))
src/omero/plugins/user.py:            self.ctx.dbg(traceback.format_exc(sv))
```

joshmoore:
Things look to be running well today in devspace. Merging.

joshmoore:
Oops:
``` 
File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero/plugins/sessions.py", line 450, in login
    % traceback.format_exception(None, e, e.__traceback__))
AttributeError: 'exceptions.IOError' object has no attribute '__traceback__'
```
```
(omero-virtualenv) bash-4.2$ python -V
Python 2.7.5```

PR #48, opened 2019-10-21
Py3 sychronization

joshmoore:
As we approach having merge-ci green, it's important to at least have tests _running_ on Python 3. This branch is intended to get py3-ci to a state where integration tests are run (without making merge-ci go red).

cc: @ome/python


joshmoore:
Note: this unit test currently does nothing because it needs `OMERODIR` set. I'll look into downloading the current state of `develop` to get us at least that far. Alternatives welcome.

PR #50, opened 2019-10-22
Use sys.exc_info rather than __traceback__

joshmoore:
see: https://github.com/ome/omero-py/pull/43


joshmoore:
Merging as a quick fix. Looks like these code paths don't have a clear integration test. :/

cc: @ome/python

PR #52, opened 2019-10-22
get-user

jburel:
handle case when last value is not set
The error occurs in Python3


joshmoore:
Which test did this fix, @jburel?

jburel:
I could not run locally in Python 3 OmeroPy.test.integration.clitest.test_user.TestUser.testList without the change
I still cannot reproduce the issue notice on CI but at least the tests were green locally

PR #53, opened 2019-10-22
Remove PYTHONPATH from tox.ini

manics:
There's no need to add `target` to `PYTHONPATH`, `python setup.py install` should do everything


joshmoore:
Thanks, @manics. Not sure where this left over from. Merging based on travis' greenness.

PR #57, opened 2019-10-23
Output

jburel:
Adjust commit introduced in https://github.com/ome/omero-py/pull/39
https://github.com/ome/omero-py/pull/39/commits/f9a77bd416721caee745abe9038f610f5a9322a0
This broke the integration tests on CI
I was able to reproduce the issue locally too

this should fix https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/194/testReport/junit/OmeroPy.test.integration.clitest.test_fs/TestFS/testRepos/
and
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/194/testReport/junit/OmeroPy.test.integration.clitest.test_fs/TestFS/testSetsWithTransfer/

cdff1dee505ce341ef6c8bd0dcb50ba5c7ac9801
should fix https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/194/testReport/junit/OmeroPy.test.integration.clitest.test_user/TestUser/testList_None_last_name_/


src/omero/util/text.py
```diff
@@ -320,7 +320,8 @@ def tostring(x):
                 try:
                     return str(x)
                 except UnicodeDecodeError:
-                    return '<Invalid UTF-8>'
+                    # Unicode characters are present
+                    return str(x.decode("utf-8", "ignore"))
```
joshmoore: I'm have some concern that this is hiding a loss of information. e.g. someone queries, get's back `flgel` and then uses that as input. If automated, `'<Invalid UTF-8>'` had the same risk, but I'm unsure of what the original intent is.

jburel: The output from the CLI command is actually ``Flügel``

manics: What about if you use `replace` instead of `ignore`?

joshmoore: > The output ... is actually Flügel

In other words, some previous bug has been remedied? If so, that's great. I wonder though if something else would get dropped.

jburel: ``replace`` or ``ignore`` returns the same output for the case tested

manics: Is this try-catch specifc to python 2 (I'm not too bothered since we know Python 2 has unicode issues) or Python 3 too?

jburel: Python 2 there is another try-catch for Python 3

joshmoore:
Only 3 unrelated integration tests left ([build 198](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/198/#showFailuresLink)). Merging.

Note: if more discussion is needed on `ignore` vs. `replace` then an issue should likely be filed.

Issue #63, opened 2019-10-24
`admin diagnostics` has errors printing under Python 3


joshmoore:
```
Server:     icegridnode                    running
Server:     b'Blitz-0'                     Error printing text
b"error: couldn't find server `bBlitz-0'"
Server:     b'DropBox'                     Error printing text
b"error: couldn't find server `bDropBox'"
Server:     b'FileServer'                  Error printing text
b"error: couldn't find server `bFileServer'"
Server:     b'Indexer-0'                   Error printing text
b"error: couldn't find server `bIndexer-0'"
Server:     b'MonitorServer'               Error printing text
b"error: couldn't find server `bMonitorServer'"
Server:     b'OMERO.Glacier2'              Error printing text
b"error: couldn't find server `bOMERO.Glacier2'"
Server:     b'OMERO.IceStorm'              Error printing text
b"error: couldn't find server `bOMERO.IceStorm'"
Server:     b'PixelData-0'                 Error printing text
b"error: couldn't find server `bPixelData-0'"
Server:     b'Processor-0'                 Error printing text
b"error: couldn't find server `bProcessor-0'"
Server:     b'Tables-0'                    Error printing text
b"error: couldn't find server `bTables-0'"
Server:     b'TestDropBox'                 Error printing text
b"error: couldn't find server `bTestDropBox'"
```

jburel:
See https://github.com/ome/omero-py/pull/65

PR #65, opened 2019-10-24
diagnotics output

jburel:
Fix the output of ``omero admin diagnostics``
Tested with both Python 2 and 3

see https://github.com/ome/omero-py/issues/63


jburel:
There is no test checking the output
This will have to be checked for now using the command itself

PR #68, opened 2019-10-25
Use ImportError instead of ModuleNotFoundError.

mtbc:
Ubuntu xenial ships with Python 3.5 so this PR helps with getting that working.

We may instead choose to require Python 3.6.

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -301,7 +301,7 @@ class BaseClient(object):
             # Register Object Factory
             try:
                 import ObjectFactoryRegistrar as ofr
-            except ModuleNotFoundError:
+            except ImportError:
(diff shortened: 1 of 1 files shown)


src/omero/clients.py
```diff
@@ -301,7 +301,7 @@ def _initData(self, id):
             # Register Object Factory
             try:
                 import ObjectFactoryRegistrar as ofr
-            except ModuleNotFoundError:
+            except ImportError:
```
manics:
```
$ python --version
Python 3.6.9 :: Anaconda, Inc.

$ python -c 'print(ModuleNotFoundError.mro())'
[<class 'ModuleNotFoundError'>, <class 'ImportError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>]
```

PR #69, opened 2019-10-27
Fix file_ann.getFileInChunks() - start with b''

will-moore:
Needed for e.g. opening Figure files.
See https://github.com/ome/omero-figure/pull/344/files#diff-0a76704154b08eff719dbe89fa944468R355

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -5303,7 +5303,7 @@ class _OriginalFileAsFileObj(object):
         return self.pos
 
     def read(self, n=-1):
-        buf = ''
+        buf = b''
         if n < 0:
             endpos = self.rfs.size()
         else:
--- a/src/omero/testlib/__init__.py
+++ b/src/omero/testlib/__init__.py
@@ -322,8 +322,8 @@ class ITest(object):
             raise Exception("import failed: [%r] %s\n%s\n%s" % (
                 args, rc, out, err))
         pix_ids = []
-        for x in out.split("\n"):
-            if x and x.find("Created") < 0 and x.find("#") < 0:
+        for x in out.split(b"\n"):
+            if x and x.find(b"Created") < 0 and x.find(b"#") < 0:
                 try:    # if the line has an image ID...
                     image_id = str(int(x.strip()))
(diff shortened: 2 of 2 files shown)


src/omero/gateway/__init__.py
```diff
@@ -5303,7 +5303,7 @@ def tell(self):
         return self.pos
 
     def read(self, n=-1):
-        buf = ''
+        buf = b''
```
jburel: I am not convinced that the strategy to follow
This could leads to problem like the one in https://github.com/ome/omero-py/pull/70

joshmoore: This is generally the problem everywhere. We now need to be certain in each method if a variable is a bytes or a str. As long as we're working in unison, i.e. agree on what's being returned where, then there's nothing inherently wrong with saying a given location is `bytes`.

jburel: Going over the method after the fact will be tricky and potentially harder to catch the issue 
Indicating the expected output needs to be mentioned as the problem is tackled otherwise we will introduce "bugs" without realising

manics: I think returning bytes is correct here since it's the contents of a file, **not** a string.

sbesson:
Agreed on collectively agreeing on the contract of this API and making sure the docstring document the return type unambiguously.

Two options I can think of to maintain consumer flexibility would be:

-  allow to pass the binary mode to constructor and consume this mode to determine the return type of `read` - this is the strategy of the standard  https://docs.python.org/3/library/functions.html#open function
- create a secondary read function with an explicit bytes/string output - this is the strategy of `file.content` and `file.content_string` of https://testinfra.readthedocs.io/en/latest/modules.html#file

joshmoore:
I could also imagine allowing the py2 and py3 versions to diverge if necessary.

joshmoore:
see also https://github.com/ome/openmicroscopy/pull/6146#issuecomment-548701610

will-moore:
Anything to change here, or just keeping this open until we've discussed strategy?

joshmoore:
Merging for 5.6.dev3

PR #70, opened 2019-10-29
Py3: upgrade message

jburel:
fix formatting error
Without the change this leads to 
```
omero admin checkupgrade
WARNING:omero.util.UpgradeCheck:b'UPGRADE AVAILABLE:Please upgrade to 5.5.1. See http://downloads.openmicroscopy.org/latest/omero for the latest version.'
Error printing text

```

No test for checking output. This was done manually

### DIFF
--- a/src/omero/util/upgrade_check.py
+++ b/src/omero/util/upgrade_check.py
@@ -153,4 +153,4 @@ class UpgradeCheck(object):
             self._set(None, None)
         else:
             self.log.warn("UPGRADE AVAILABLE:" + result.decode('utf-8'))
-             self._set(result.decode('utf-8'), None)
+            self._set(result.decode('utf-8'), None)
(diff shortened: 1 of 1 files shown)


src/omero/util/upgrade_check.py
```diff
@@ -152,5 +152,5 @@ def run(self):
             self.log.info("no update needed")
             self._set(None, None)
         else:
-            self.log.warn(b"UPGRADE AVAILABLE:" + result)
+            self.log.warn("UPGRADE AVAILABLE:" + result.decode('utf-8'))
             self._set(result, None)
```
sbesson:
```suggestion
             self._set(result.decode('utf-8'), None)
```

sbesson:
Looks like b556c077469926fce889a44e75c59cf1bf372002 introduced a trailing space indent probably arising my inline suggestion.

PR #73, opened 2019-10-30
Py3 admin

jburel:
This PR should fix test like 
* https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/27/testReport/junit/OmeroPy.test.integration.clitest.test_pyramids/TestRemovePyramidsFullAdmin/test_remove_pyramids_little_endian/
Also fixed some remaining flake8 issues

### DIFF
--- a/src/omero/plugins/admin.py
+++ b/src/omero/plugins/admin.py
@@ -1020,8 +1020,11 @@ present, the user will enter a console""")
         from omero.util.cleanse import removepyramids
         client = self.ctx.conn(args)
         client.getSessionId()
-        wait = args.wait if args.wait > 0 else 25
-        limit = args.limit if args.limit > 0 else 500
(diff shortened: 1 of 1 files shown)


src/omero/plugins/admin.py
```diff
@@ -1020,8 +1019,11 @@ def removepyramids(self, args):
         from omero.util.cleanse import removepyramids
         client = self.ctx.conn(args)
         client.getSessionId()
-        wait = args.wait if args.wait > 0 else 25
-        limit = args.limit if args.limit > 0 else 500
+        wait = args.wait if args.wait is not None and int(args.wait) > 0 else 25
```
sbesson: I am largely sure this check was not necessary since unlike the `limit` argument, the `wait` argument has a default value and should never by None. It does not hurt in all cases.

PR #80, opened 2019-10-30
Test py 3.7 in travis

manics:
We officially support Python 3.6 but since we'll probably support 3.7 at some point in the future we might as well ensure the unit tests pass.

See https://github.com/ome/zeroc-ice-py-manylinux/pull/2 to speed this up

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27, py36
+envlist = py27, py36, py37
 # https://tox.readthedocs.io/en/latest/config.html#conf-requires
 # Ensure pip is new so we can install manylinux wheel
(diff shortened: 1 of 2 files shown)


joshmoore:
- Run without wheel:  14 min 56 sec
 - Run with wheel: 5 min 18 sec
 - Previously: ~3 min 50 sec

i.e. now there's a linear increase of about 1 minute. Considering that worth it.

tox.ini
```diff
@@ -14,11 +14,13 @@ deps =
     PyYAML
     py27: tables < 3.6.0
     py36: tables
+    py37: tables
```
joshmoore: There are ways to glob to prevent repeating, but I assume we can dig those out later.

PR #86, opened 2019-10-31
plugins: obj

jburel:
Message no longer exists
Print the error instead

* https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/29/testReport/OmeroPy.test.integration.clitest.test_obj/TestObj/test_get_unknown_and_empty_field/
* https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/29/testReport/OmeroPy.test.integration.clitest.test_obj/TestObj/test_list_get/

### DIFF
--- a/src/omero/plugins/obj.py
+++ b/src/omero/plugins/obj.py
@@ -150,7 +150,7 @@ class TxAction(object):
         parts = self.tx_cmd.type.split(":")
         try:
             return int(parts[1])
-        except:
+        except Exception:
             return None
 
     def instance(self, ctx):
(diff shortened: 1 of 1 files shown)


joshmoore:
Is there a change in the output for the user? One option would be to access the message via `args`: https://portingguide.readthedocs.io/en/latest/exceptions.html#iterating-exceptions

joshmoore:
https://github.com/ome/openmicroscopy/pull/6144/files#diff-963231c1845e662ac0e67d218bcf118eR102 suggests that there is _no_ change in the output?

jburel:
I did not notice any changes. I could use ``args`` as suggested in the guide if we prefer

jburel:
Read message from args as suggested in the porting guide

jburel:
The 2 tests listed in the description should still be green in both py2 and 3

joshmoore:
Both tests pass in merge (224) and py3 (40)

PR #88, opened 2019-10-31
Fix fix_clitest.test_download handle std.out

will-moore:
Fixes https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/29/testReport/OmeroPy.test.integration.clitest.test_download/TestDownload/testOriginalFileStdout__/

### DIFF
--- a/src/omero/plugins/download.py
+++ b/src/omero/plugins/download.py
@@ -36,6 +36,20 @@ Examples:
     bin/omero download Image:5 original_image
 """
 
+class StdOutHandle():
+    """
+    File handle for writing bytes to std.out in python 2 and python 3
+    """
+    # https://github.com/pexpect/pexpect/pull/31/files
+    @staticmethod
+    def write(b):
+        # Handle stdout.write for bytes
+        try:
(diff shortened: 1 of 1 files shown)


src/omero/plugins/download.py
```diff
@@ -37,6 +37,22 @@
...
+    File handle for writing bytes to std.out in python 2 and python 3
+    """
+    # https://github.com/pexpect/pexpect/pull/31/files
+    @staticmethod
+    def write(b):
+        # Handle stdout.write for bytes
+        try:
+            # Try writing bytes... python 2
+            return sys.stdout.write(b)
+        except TypeError:
```
joshmoore: Do you know why is this a TypeError rather than an AttributeError in the example?

will-moore: Because the ```write(b)``` is expecting a String in python3, so ```bytes``` is the wrong type.

joshmoore: Ah, ok. If it had been written Python 3 first, it would be an AttributeError under Python 2.

Issue #84, opened 2019-10-31
 `admin diagnostics` has errors printing under Python 3 (2)


sbesson:
Follow-up of #63 

```
00:49:52 b"error: couldn't find application `bOMERO'"
```

the SSL/TCP ports reporting might need to be fixed similarly to #65

PR #94, opened 2019-10-31
Output diagnostics 2

jburel:
This PR fixes the remaining issues when determining the port used
It also prints out the value of the new env variable ``OMERODIR``
See https://github.com/ome/omero-py/issues/84


src/omero/plugins/admin.py
```diff
@@ -1449,13 +1450,14 @@ def port_val(port_type, value):
                     self._item("%s %s port" % (s, port_type),
                                "%s" % value or "Not found")
                     self.ctx.out("")
+                s = str(s, "utf-8")
```
sbesson: I did not know about this variant to `bytes.decode()` which I find quite attractive. Only question is whether we are okay with a mixture of styles across our code or whether it is preferable to unify on one way to decoding bytes.

jburel:
@sbesson as discussed this morning I have used ``decode``.
Could you check again?

PR #100, opened 2019-11-04
Handle getObjects(attributes={id:ID} as long

will-moore:
Fixes many 'show' tests, which use
```getObjects('Image', attributes={'id':1})```
We handle ```id``` to wrap in ```rlong```.
E.g: - https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/37/testReport/OmeroWeb.test.integration.test_show/TestShow/test_project_legacy_path/

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -3407,7 +3407,12 @@ class _BlitzGateway (object):
                     rv = rlong(v)
                 else:
                     # lookup type of attribute from class
-                    klass = wrapper.OMERO_TYPE
+                    if wrapper.OMERO_CLASS is not None:
+                        klass = getattr(omero.model,
+                                        "%sI" % wrapper.OMERO_CLASS)
+                    else:
+                        # AnnotationWrappers don't have OMERO_CLASS
+                        klass = wrapper.OMERO_TYPE
                     rtype = getattr(klass._field_info, k).wrapper(1)
                     rv = rtype.__class__(v)
                 baseParams.map[k] = rv
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -3405,8 +3405,8 @@ class _BlitzGateway (object):
                 clauses.append('obj.%s=:%s' % (k, k))
                 if k == 'id':
                     rv = rlong(v)
-                else:
-                    # lookup type of attribute from class
+                elif isinstance(v, IntType):
+                    # lookup rint vv rlong from class
                     if wrapper.OMERO_CLASS is not None:
                         klass = getattr(omero.model,
                                         "%sI" % wrapper.OMERO_CLASS)
@@ -3415,6 +3415,8 @@ class _BlitzGateway (object):
                         klass = wrapper.OMERO_TYPE
                     rtype = getattr(klass._field_info, k).wrapper(1)
                     rv = rtype.__class__(v)
+                else:
+                    rv = omero_type(v)
                 baseParams.map[k] = rv
         if clauses:
             query += " where " + (" and ".join(clauses))
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -3413,7 +3413,7 @@ class _BlitzGateway (object):
                     else:
                         # AnnotationWrappers don't have OMERO_CLASS
                         klass = wrapper.OMERO_TYPE
-                    rtype = getattr(klass._field_info, k).wrapper(1)
+                    rv = getattr(klass._field_info, k).wrapper(v)
                     rv = rtype.__class__(v)
                 else:
                     rv = omero_type(v)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/omero/gateway/__init__.py [hunk 1]
```diff
@@ -3403,7 +3403,11 @@ def buildQuery(self, obj_type, ids=None, params=None, attributes=None,
         if attributes is not None:
             for k, v in list(attributes.items()):
                 clauses.append('obj.%s=:%s' % (k, k))
-                baseParams.map[k] = omero_type(v)
+                if k == 'id':
```
joshmoore: Does this mean that for other very large values (timestamps?) that they will get converted into `rints` and possibly overflow?

will-moore: The only 'attributes' we support here are direct attributes of the objects supported by ```getObjects()```:
```
"project": ProjectWrapper,
                           "dataset": DatasetWrapper,
                           "image": ImageWrapper,
                           "screen": ScreenWrapper,
                           "plate": PlateWrapper,
                           "plateacquisition": PlateAcquisitionWrapper,
                           "acquisition": PlateAcquisitionWrapper,
                           "well": WellWrapper,
                           "roi": RoiWrapper,
                           "shape": ShapeWrapper,
                           "experimenter": ExperimenterWrapper,
                           "experimentergroup": ExperimenterGroupWrapper,
                           "originalfile": OriginalFileWrapper,
                           "fileset": FilesetWrapper,
                           "commentannotation": CommentAnnotationWrapper,
                           "tagannotation": TagAnnotationWrapper,
                           "longannotation": LongAnnotationWrapper,
                           "booleanannotation": BooleanAnnotationWrapper,
                           "fileannotation": FileAnnotationWrapper,
                           "doubleannotation": DoubleAnnotationWrapper,
                           "termannotation": TermAnnotationWrapper,
                           "timestampannotation": TimestampAnnotationWrapper,
                           "mapannotation": MapAnnotationWrapper,
```
I guess I could see ```longannotation.value```, but are there any others?

joshmoore:
```
Image.acquisitionDate
Pixels.timeIncrement
TimestampeAnnotation.value
PlateAcquisition.{start,end}Time
WellSample.timepoint
OriginalFile.size
```

will-moore: I could try:
```
if k in ('id', 'acquisitionDate',  'timeIncrement', 'startTime', 'endTime', 'timepoint', 'size':
    rlong(v)
```
but I didn't include 'value' since that could be any text/int value.
But, do we really want to support all these, write tests etc and update the list if we support more objects? I struggle to imagine use cases for finding these objects via these attributes.
We only really use ```attributes``` for name, or tag textValue etc.

joshmoore: Would it be an option at this point to use something like:

```
In [14]: getattr(omero.model.OriginalFileI._field_info, "size").wrapper(1)
Out[14]:
object #0 (::omero::RLong)
{
    _val = 1
}
```

will-moore:
Those commits caused failure at https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/48/testReport/OmeroPy.test.integration.gatewaytest.test_get_objects/TestFindObject/testFindAnnotation/

joshmoore:
Then `omero_type` is doing something more. `wrapper(1)` gives you an `omero.RLong`. How would you convert that to whatever has an `OMERO_CLASS`?

will-moore:
This was failing:
```
>                   if wrapper.OMERO_CLASS is not None:
E                   AttributeError: 'function' object has no attribute 'OMERO_CLASS'
```
because ```getObject("Annotation")``` has an abstract ```AnnotationWrapper``` class.
It should be fixed by my last commit above, which only uses the new behaviour when needed (when we need to distinguish between rint and rlong)

will-moore:
NB: I still had to handle ```id``` on it's own since ```id``` is not included in the ```_field_info```.

src/omero/gateway/__init__.py
```diff
@@ -3403,7 +3403,21 @@ def buildQuery(self, obj_type, ids=None, params=None, attributes=None,
...
+                    rv = rlong(v)
+                elif isinstance(v, IntType):
+                    # lookup rint vv rlong from class
+                    if wrapper.OMERO_CLASS is not None:
+                        klass = getattr(omero.model,
+                                        "%sI" % wrapper.OMERO_CLASS)
+                    else:
+                        # AnnotationWrappers don't have OMERO_CLASS
+                        klass = wrapper.OMERO_TYPE
+                    rtype = getattr(klass._field_info, k).wrapper(1)
```
joshmoore:
```suggestion
                    rv = getattr(klass._field_info, k).wrapper(v)
```

src/omero/gateway/__init__.py
```diff
@@ -3403,7 +3403,21 @@ def buildQuery(self, obj_type, ids=None, params=None, attributes=None,
...
+                    rv = rtype.__class__(v)
```
joshmoore:
```suggestion
```

PR #105, opened 2019-11-06
Convert tables v1 version to string

manics:
This may or may not fix ome/openmicroscopy#6114 (comment)

Moved from https://github.com/joshmoore/omero-py/pull/14


joshmoore:
merge 228 and py3 46 look good. No table version errors.

PR #106, opened 2019-11-06
fix testDeleteObjectDirect()

will-moore:
Fixes new test: https://github.com/ome/openmicroscopy/pull/6170

Not sure how / why /when this has started failing.


joshmoore:
Change looks good and added test is passing :+1:

PR #107, opened 2019-11-06
Include python executable parent PATH in call()

manics:
This ensures that if virtualenv python is executed directly without activating the virtualenv the virtualenv bin PATH is included in the search PATH.

For example
`OMERODIR=/tmp/omeroweb /path/to/venv/bin/omero web start --foreground` should work even if you haven't activated the virtualenv. Previously this would fail as `/path/to/venv/bin/gunicorn` wouldn't be found.

### DIFF
--- a/src/omero/cli.py
+++ b/src/omero/cli.py
@@ -1388,7 +1388,7 @@ class CLI(cmd.Cmd, Context):
 
     def _env(self):
         """
-        Configure environment with PYTHONPATH as
+        Configure environment with PYTHONPATH and PATH as
         setup by bin/omero
 
         This list needs to be kept in line with OmeroPy/bin/omero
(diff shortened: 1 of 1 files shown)


src/omero/cli.py
```diff
@@ -1409,6 +1409,16 @@ def _env(self):
             else:
                 pypath = "%s%s%s" % (pypath, os.path.pathsep, paths)
         env["PYTHONPATH"] = pypath
+
+        ospath = env.get("PATH", None)
+        binpypath = os.path.dirname(sys.executable)
+        if ospath:
+            if binpypath not in os.path.split(os.path.pathsep):
```
joshmoore: I imagine there are cases that this would not detect (symlinks, relative paths), but the worst that can happen is that a directory gets multiply listed.

manics:
Note this isn't as beneficial as I anticipated. You still need to set `OMERODIR` for `omero web` so arguably setting `PATH` and/or activating the virtualenv is not that big a deal. Still arguably useful though?

joshmoore:
Seems fine. :+1:

Issue #109, opened 2019-11-07
omero db script fails if LANG is not set


manics:
```
-bash-4.2$ echo $LANG

-bash-4.2$ OMERODIR=/opt/omero/server/OMERO.server/ /opt/omero/server/venv3/bin/omero db script -f /tmp/omero.sql "" "" omero
WARNING: Positional arguments are deprecated
Using OMERO5.4 for version
Using 0 for patch
Using password from commandline
Traceback (most recent call last):
  File "/opt/omero/server/venv3/bin/omero", line 131, in <module>
    rv = omero.cli.argv()
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1746, in argv
    cli.invoke(args[1:])
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1187, in invoke
    stop = self.onecmd(line, previous_args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1264, in onecmd
    self.execute(line, previous_args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/cli.py", line 1346, in execute
    args.func(args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/install/windows_warning.py", line 26, in wrapper
    return func(self, *args, **kwargs)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/plugins/db.py", line 315, in script
    self._create(sql, db_vers, db_patch, pwhash, args)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/plugins/db.py", line 222, in _create
    self._copy(header, output, str, cfg)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/plugins/db.py", line 150, in _copy
    for s in input:
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 2124: ordinal not in range(128)
```
Expected: `omero db` should treat everything as UTF-8

PR #110, opened 2019-11-07
db: read and write utf-8

manics:
The database SQL files are known to be in UTF-8, so don't rely on `LANG` being set to `utf-8`
This relies on the native argparse library.

Closes https://github.com/ome/omero-py/issues/109

## Testing:
- stdout: `LANG= OMERODIR=~/tmp/OMERO.server omero db script "" "" omero -f -`
- default output: `LANG= OMERODIR=~/tmp/OMERO.server omero db script "" "" omero`
- named file output: `LANG= OMERODIR=~/tmp/OMERO.server omero db script "" "" omero -f output.sql`

### DIFF
--- a/src/omero/plugins/db.py
+++ b/src/omero/plugins/db.py
@@ -145,7 +145,7 @@ class DatabaseControl(BaseControl):
         return value.strip()
 
     def _copy(self, input_path, output, func, cfg=None):
-            input = open(str(input_path))
+            input = open(str(input_path), encoding='utf-8')
             try:
                 for s in input:
                         try:
@@ -206,7 +206,7 @@ class DatabaseControl(BaseControl):
         else:
             script = "%s__%s.sql" % (db_vers, db_patch)
             location = old_div(path.getcwd(), script)
-            output = open(location, 'w')
+            output = open(location, 'w', encoding='utf-8')
             self.ctx.out("Saving to " + location)
 
         try:
--- a/src/omero/plugins/db.py
+++ b/src/omero/plugins/db.py
@@ -150,7 +150,10 @@ class DatabaseControl(BaseControl):
         return value.strip()
 
     def _copy(self, input_path, output, func, cfg=None):
-            input = open(str(input_path), encoding='utf-8')
+            try:
+                input = open(str(input_path), encoding='utf-8')
+            except TypeError:
+                input = open(str(input_path))
             try:
                 for s in input:
                         try:
@@ -211,7 +214,10 @@ class DatabaseControl(BaseControl):
         else:
             script = "%s__%s.sql" % (db_vers, db_patch)
             location = old_div(path.getcwd(), script)
-            output = open(location, 'w', encoding='utf-8')
+            try:
+                output = open(location, 'w', encoding='utf-8')
+            except TypeError:
+                output = open(location, 'w')
             self.ctx.out("Saving to " + location)
 
         try:
--- a/src/omero/plugins/db.py
+++ b/src/omero/plugins/db.py
@@ -33,7 +33,7 @@ from past.utils import old_div
 from omero.cli import BaseControl
 from omero.cli import CLI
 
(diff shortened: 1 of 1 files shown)


manics:
This is needed for the Ansible role: https://github.com/ome/ansible-role-omero-server/pull/40/files#diff-e23e5ddc5344dbaf4fa8f3d569c0f005R140

sbesson:
I was unable to reproduce the previous issue locally (by downloading the latest OMERO build, unsetting LANG and using either of the commands above). The changes generally look sensible. I can try and retest in an environment where the original bug can be reproduced today.

src/omero/plugins/db.py
```diff
@@ -33,7 +33,7 @@
 from omero.cli import BaseControl
 from omero.cli import CLI
 
-from omero_ext.argparse import FileType, SUPPRESS
+from argparse import FileType, SUPPRESS
```
sbesson: This causes the help to be printed as follows:

```
[omero@ac72d3116bac tmp]$ venv2/bin/omero db script --help
ERROR:omero.gateway:No Pillow installed, line plots and split channel will fail!
usage: venv2/bin/omero db script [-h] [-f FILE] [--version VERSION]
                                 [--patch PATCH] [--password PASSWORD]
                                 [--no-salt]
                                 [posversion] [pospatch] [pospassword]

Positional Arguments:
  posversion            ==SUPPRESS==
  pospatch              ==SUPPRESS==
  pospassword           ==SUPPRESS==

Optional Arguments:
  In addition to any higher level options

  -h, --help            show this help message and exit
  -f FILE, --file FILE  Optional file to save to. Use '-' for stdout.
  --version VERSION     ==SUPPRESS==
  --patch PATCH         ==SUPPRESS==
  --password PASSWORD   OMERO root password
  --no-salt             Disable the salting of passwords
```

Does this mean we need to switch everything to the upstream `argparse` to make things work?

manics: https://docs.python.org/2/library/argparse.html#help
> argparse supports silencing the help entry for certain options, by setting the help value to argparse.SUPPRESS

If it's due to an interaction between the distribution and omero_ext versions it's probably necessary to drop `omero_ext.argparse`, especially as we no longer support Python 2.6

manics:
`argparse.FileType` on Python 2 does not support the encoding parameter: https://docs.python.org/2/library/argparse.html#filetype-objects so I'm don't know whether it's possible to fix stdout.

sbesson:
> argparse.FileType on Python 2 does not support the encoding parameter: https://docs.python.org/2/library/argparse.html#filetype-objects so I don't know whether it's possible to fix stdout.

Makes sense for Python 2. On Python 3, is there a reason why we could not fix stdout?

> If it's due to an interaction between the distribution and omero_ext versions it's probably necessary to drop omero_ext.argparse, especially as we no longer support Python 2.6

Agreed, in order to fix this issue, it looks like the only viable options are either to  switch to the upstream `argparse` across the code or to backport the relevant commits on our copy. Do you want to push another commit achieving the former that can be validate against the integration test?

manics:
> Makes sense for Python 2. On Python 3, is there a reason why we could not fix stdout?

It's working for me on Python 3!

sbesson:
After additional debugging on the stdout case in the CentOS based environment described above, I can confirm `LANG= venv1/bin/omero db script "" "" omero -f -` creates a `<_io.TextIOWrapper name='<stdout>' mode='w' encoding='ANSI_X3.4-1968'>` object. However, `LANG=en_GB.UTF-8 venv1/bin/omero db script "" "" omero -f -` works.

In general if writing to stdout, expeting `LANG` to be set to a proper UTF-8 environment while writing the output of the SQL scripts is probably an acceptable assumption. I will try and investigate this scenario separately.

As https://github.com/ome/omero-py/pull/115 was opened separately to investigate the `argparse` switch, merging this PR  for a dev release and the discussion on the `argparse` library can happen in #115.

PR #114, opened 2019-11-11
First attempt at adding optional packages

manics:
Can be tested locally:
- `pip install -e '.[extra]'`
- `pip install -e '.[linux]'`
- New: `pip install -e '.[testlib]'`

The general idea is to manage the packages installed by the Ansible role here to reduce the number of time dependencies are defined or pinned.


manics:
Closing this, it's out of date and at some point omero-py will be split up further

PR #115, opened 2019-11-12
Remove omero_ext.argparse

manics:
See https://github.com/ome/omero-py/pull/110

~~This will conflict~~ rebased

--exclude see https://github.com/ome/omero-py/pull/115#pullrequestreview-322991815


joshmoore:
```
E           Exception: Missing prompt: 'Server: [localhost:4064]'
```
etc.

manics:
I'm still no nearer to finding out what's happening.

sbesson:
The last two commits fix the upstream conflicts following the inclusion of #123 and the failing unit tests (33f0017). It looks like the upstream `argparse` enforces positional arguments to be appended to the command.

This should be ready for re-inclusion in the daily builds and a round of CLI review.

In terms of API, is there a way to have `omero_ext.argparse` simply alias the upstream `argparse` library and deprecate it?

joshmoore:
> It looks like the upstream `argparse` enforces positional arguments

This will have a pretty wide impact. e.g. `bin/omero -s localhost login` is pretty common.

sbesson:
> This will have a pretty wide impact. e.g. bin/omero -s localhost login is pretty common.

Agreed, I am pretty certain all the CLI integration tests are appending the command to the login arguments for instance. Probably worth looking into an option to maintain some backwards compatibility.

PR #125, opened 2019-11-20
changelog: list breaking changes

joshmoore:
- [x] getFileInChunks
 - [x] exportOmeTiff
 - [x] BytesIO etc.
 - [x] rlongs

cc: @ome/python

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 
 - Fixes primarily focused on byte return types
 
+## API Changes
+
+- getFileInChunks now returns a bytes-generator
+
 # 5.6.dev2 (October 2019)
 
 - Even more Python 3 fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@
 
 ## API Changes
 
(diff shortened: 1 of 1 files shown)


will-moore:
```image.exportOmeTiff()```

**Comment by will-moore** [comment]
Previously where you did:
```
image = conn.getObject("Image", id)
Image.open(StringIO(image.renderJpeg))
```
now you need to do:
```
Image.open(BytesIO(image.renderJpeg))
```
That's not so much a BlitzGateway change, since it's just returning ```re.renderCompressed()```

joshmoore:
Agreed. I think we add BytesIO, `str.decode()` etc to a more general migration guide as opposed to our own CHANGELOG.

will-moore:
```conn.createOriginalFileFromFileObj()``` needs to take ```BytesIO``` file object instead of StringIO.

will-moore:
Do we want to mention ```omero_type(longValue) and omero.rtypes.wrap(longValue)``` won't return ```rlong```?

joshmoore:
Merging after a quick sign-off for posting on image.sc. Happy to have follow-ups. thanks, @will-moore.

PR #127, opened 2019-11-20
batch annotate file fix

will-moore:
Fixes bug reported on Row 14 of https://docs.google.com/spreadsheets/d/1ry6vS0dmsylFQDAiIZDrPC3kXNpnHSMtK-2yj5X8jaU/edit#gid=420881033
"Multi-select images, then try to upload an attachment..."

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -3668,7 +3668,7 @@ class _BlitzGateway (object):
         else:
             # for multiple parents, we first need to find annotations linked to
             # ALL of them, then exclude those from query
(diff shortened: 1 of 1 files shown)


manics:
Code looks fine. on py3-ci:
1. Select multiple images
2. Attach a file (existing or upload a new one)
3. After it's attached click on `+` again. this causes a feedback error to popup (seems to be related to having existing file annotations?)

will-moore:
@manics Good catch. I'd actually found that same error when writing tests for this at https://github.com/ome/openmicroscopy/pull/6182 and had fixed it locally but forgot to commit.

pwalczysko:
Fix confirmed on py3-ci and merge-ci

PR #130, opened 2019-11-26
config_parser: further futurize

joshmoore:
Necessary for work on the documentation merge build:
https://github.com/ome/ome-documentation/pull/2035

### DIFF
--- a/src/omero/install/config_parser.py
+++ b/src/omero/install/config_parser.py
@@ -343,7 +343,7 @@ class PropertyParser(object):
         from django.conf import settings
 
         for key, values in sorted(
-                iter(settings.CUSTOM_SETTINGS_MAPPINGS.items()),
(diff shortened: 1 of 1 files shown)


src/omero/install/config_parser.py
```diff
@@ -372,8 +372,8 @@ def headers(self):
                     additional_headers[section] = Header(section)
                 headers.setdefault(additional_headers[section], []).append(x)
 
-        for key in headers.keys():
-            headers[key].sort(lambda a, b: cmp(a.key, b.key))
+        for key in list(headers.keys()):
```
will-moore: Don't need ```list``` here really?

joshmoore: Hmmm.... pretty sure this is coming from the call to `futurize -w -0`, i.e. being overly cautious. Is it worth pushing away?

will-moore: Probably not worth the trouble.

PR #131, opened 2019-11-27
use tox-travis

joshmoore:
see: https://tox-travis.readthedocs.io/en/stable/


joshmoore:
Doesn't look to be advantageous.

PR #133, opened 2019-11-28
Add unit test covering Unicode

sbesson:
See also https://github.com/ome/openmicroscopy/pull/6189

b45e95c exposes a Python 3.6 regression when adding a StringColumn containing Unicode. The same scenario passes without issue on Python 2

### DIFF
--- a/test/unit/tablestest/test_hdfstorage.py
+++ b/test/unit/tablestest/test_hdfstorage.py
@@ -247,14 +247,14 @@ class TestHdfStorage(TestCase):
         cols = [omero.columns.StringColumnI("name", "description", 16, None)]
         hdf.initialize(cols)
         cols[0].settable(hdf._HdfStorage__mea)  # Needed for size
-        cols[0].values = ["foo", "αbc"]
+        cols[0].values = ["foo", "მიკროსკოპის პონი"]
         hdf.append(cols)
         rows = hdf.getWhereList(time.time(), '(name=="foo")', None, 'b', None,
                                 None, None)
         assert 1 == len(rows)
         assert 16 == hdf.readCoordinates(time.time(), [0],
                                          self.current).columns[0].size
-        rows = hdf.getWhereList(time.time(), '(name=="αbc")', None, 'b', None,
+        rows = hdf.getWhereList(time.time(), '(name=="მიკროსკოპის პონი")', None, 'b', None,
                                 None, None)
         assert 1 == len(rows)
         assert 16 == hdf.readCoordinates(time.time(), [0],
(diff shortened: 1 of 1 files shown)


test/unit/tablestest/test_hdfstorage.py
```diff
@@ -247,11 +247,16 @@ def testStringCol(self):
         cols = [omero.columns.StringColumnI("name", "description", 16, None)]
         hdf.initialize(cols)
         cols[0].settable(hdf._HdfStorage__mea)  # Needed for size
-        cols[0].values = ["foo"]
+        cols[0].values = ["foo", "αbc"]
```
manics: Nitpicky suggestion: use characters that are not a homoglyph so it's clear to others?
```suggestion
        cols[0].values = ["foo", "მიკროსკოპის პონი"]
```

test/unit/tablestest/test_hdfstorage.py
```diff
@@ -247,11 +247,16 @@ def testStringCol(self):
...
         hdf.append(cols)
         rows = hdf.getWhereList(time.time(), '(name=="foo")', None, 'b', None,
                                 None, None)
         assert 1 == len(rows)
+        assert 16 == hdf.readCoordinates(time.time(), [0],
+                                         self.current).columns[0].size
+        rows = hdf.getWhereList(time.time(), '(name=="αbc")', None, 'b', None,
```
manics:
```suggestion
        rows = hdf.getWhereList(time.time(), '(name=="მიკროსკოპის პონი")', None, 'b', None,
```

test/unit/tablestest/test_hdfstorage.py
joshmoore: I think we should add `მიკროსკოპის პონი` as a constant to `omero/Constants.ice`.

sbesson:
The [numpy.dtype note](https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html?highlight=dtype%20unicode#string-dtype-note) about using strings in Python 3 is probably relevant to the root of this problem. Unfortunately, local attempts to migrate`StringColumn.dtypes()` from `S` to `U` have been unsuccessful.

Earlier demo on IDR upgraded to an experimental Python 3 environment seems to suggest that the reading of `StringColumn` created on Python 2 with Unicode characters is unaffected:

Screen Shot 2019-12-02 at 16 19 22

I expect I will not be in capacity to provide a fix for this regression for the OMERO 5.6.0. There is a question of whether this should be marked as a blocker for GA, it is certainly one for the upgrade of IDR to Python 3 as it breaks the annotation workflows if CSV files contains Unicode characters.

As immediate next steps, proposing to:
- potentially extract the Unicode bit of this test, mark it as `xfail`
- apply the same approach to the integration test in ome/openmicroscopy#6189 
- get this merged and capture the regression as an issue issue to be reviewed for OMERO 5.6.0

Alternated thoughts or suggestions welcome /cc @joshmoore @jburel @manics

joshmoore:
Something I haven't really considered yet: would a `UnicodeColumn` be of use?

manics:
What would be the difference between a `StringColumn` with unicode and a `UnicodeColumn`?

joshmoore:
It would be a location that could different read/write logic if that would help.

sbesson:
Superseded by #143

PR #134, opened 2019-11-28
travis jobs: enable Windows

joshmoore:
Follow on to #132. Run Windows in Travis.


joshmoore:
All platforms are running (periodically with tables failures) and a real issue with the Windows test arose:

```
>           return open(string, self._mode)
E           FileNotFoundError: [Errno 2] No such file or directory: 'C:UserstravisAppDataLocalTemppytest-of-unknownpytest-0popen-gw3testLoad0test'
```

joshmoore:
Green!

.travis.yml
```diff
@@ -1,41 +1,48 @@
...
+        - tox -e py37
+    - name: Python36
+      os: trusty
+      python: "3.6"
+      script:
+        - tox -e py36
+    - stage: deploy
+      os: trusty
+      python: "3.6"
+      script:
```
manics: Can you remove this `script` section? It wasn't in the original deploy step.

joshmoore: Isn't that the equivalent of the `$TOX = no` clause?

manics: `skip_cleanup` wasn't set, so it should've been deployed from a clean environment: https://docs.travis-ci.com/user/deployment#uploading-files-and-skip_cleanup

joshmoore: Hmmm.... I'm not following, but it looks like all the rst-lint actions are now in tox anyway.

joshmoore:
Few commits pushed, @manics.

manics:
Might need `script: skip` too

joshmoore:
That's us back green again, @manics. Not sure why both skips are needed.

joshmoore:
` ¯\_(ツ)_/¯` Note: this may have further issues when it comes time to deploy a tag, but I'll deal with that during release.

PR #136, opened 2019-12-02
remove hdfstorageV1 (fix #40)

### LINKED ISSUE #40: drop hdfstoragev1
see: https://github.com/ome/omero-py/pull/37

hdfstoragev2 was introduced to support breaking changes in pytables. pytables 3.6.0 (released yesterday) drops support for python 2 and so v1 is no longer needed.

joshmoore:
With the dropping of Python 2, there's no need for
a version of tables with the `openFile` method.


joshmoore:
Cheers. Debated stripping out the logic for choosing between the versions, but I figured a V3 might come along some day.

Issue #120, opened 2019-11-14
omero load directory


manics:
This script is used by the omero-server and omero-web ansible roles: https://github.com/ome/ansible-role-omero-server/blob/2.1.1/templates/omero-server-config-update-sh.j2
```
for f in {{ omero_server|web_basedir }}/config/*.omero; do
    bin/omero load "$f"
done
```
This could be built in to `omero load` instead.

PR #137, opened 2019-12-02
omero load --glob

manics:
Closes https://github.com/ome/omero-py/issues/120

For example, `omero load --glob /opt/omero/server/config/*.omero` will sort files `/opt/omero/server/config/*.omero` before loading them

### DIFF
--- a/src/omero/plugins/basics.py
+++ b/src/omero/plugins/basics.py
@@ -20,11 +20,10 @@
 from __future__ import print_function
 
 from past.builtins import cmp
+from glob import glob
 import sys
 
 from collections import defaultdict
-
-from omero_ext.argparse import FileType
 
 from omero.cli import BaseControl
 from omero.cli import CLI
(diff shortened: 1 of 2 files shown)


joshmoore:
LGTM. Merging. The only feature I can imagine is that the commands that are being loaded could also show up under `self.ctx.dbg`:

```
(omero-virtualenv) bash-4.2$ head t/0*
==> t/01.omero <==
version

==> t/02.omero <==
admin ice server pid Blitz-0

==> t/03.omero <==
config get
(omero-virtualenv) bash-4.2$ bin/omero -d1 load --glob t/*.omero
Loading file t/01.omero
5.6.dev8
Loading file t/02.omero
Executing: ['icegridadmin', '--Ice.Config=/home/omero/workspace/OMERO-test-integration/src/dist/etc/internal.cfg', '-e', 'server pid Blitz-0']
36717
Loading file t/03.omero
omero.data.dir=/home/omero/workspace/OMERO-test-integration/data
```

PR #140, opened 2019-12-04
Disable new testLoadGlob when OMERODIR not set

manics:
https://github.com/ome/omero-py/pull/14 and https://github.com/ome/omero-py/pull/137 overlapped so were not tested together.

### DIFF
--- a/test/unit/clitest/test_basics.py
+++ b/test/unit/clitest/test_basics.py
@@ -23,12 +23,14 @@
 from __future__ import unicode_literals
 from builtins import object
 import pytest
+import os
 from omero.cli import CLI
 
 cli = CLI()
 cli.loadplugins()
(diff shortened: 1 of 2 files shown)


joshmoore:
Merging proactively so we can get other PRs green, but I think a test method (additional or otherwise) that either doesn't require OMERODIR to be set (using `version`, etc) or creates its own OMERODIR would be appropriate.

manics:
I tried that in https://github.com/ome/omero-py/pull/137/commits/d1f4dc2bf259ccf3540e38a17f0cf9a7a79fda33 but `OMERODIR` needs to be set before importing the CLI.

joshmoore:
Not sure if that should be considered a bug, but you can also reset the value:

```
self.ctx.dir = NEW_OMERODIR
```

PR #141, opened 2019-12-04
Reenable testLoadGlob

manics:
Follow-up from https://github.com/ome/omero-py/pull/140#issuecomment-561652861

### DIFF
--- a/test/unit/clitest/test_basics.py
+++ b/test/unit/clitest/test_basics.py
@@ -71,7 +71,6 @@ class TestBasics(object):
         (tmp_path / 'etc').mkdir()
         (tmp_path / 'etc' / 'grid').mkdir()
         monkeypatch.setenv('OMERODIR', str(tmp_path))
-        monkeypatch.setattr(cli, 'dir', str(tmp_path))
         for i in 'abc':
             (tmp_path / (i + 'a.omero')).write_text(
                 'config set {i} {i}'.format(i=i))
--- a/test/unit/clitest/test_basics.py
+++ b/test/unit/clitest/test_basics.py
@@ -23,14 +23,12 @@
 from __future__ import unicode_literals
 from builtins import object
 import pytest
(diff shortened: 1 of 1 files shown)


test/unit/clitest/test_basics.py
```diff
@@ -69,14 +67,16 @@ def testQuit(object):
     def testVersion(object):
         cli.invoke(["version"], strict=True)
 
-    @pytest.mark.skipif(OMERODIR is False, reason="We need $OMERODIR")
-    def testLoadGlob(object, tmp_path, capsys):
+    def testLoadGlob(object, monkeypatch, tmp_path, capsys):
+        (tmp_path / 'etc').mkdir()
+        (tmp_path / 'etc' / 'grid').mkdir()
+        monkeypatch.setenv('OMERODIR', str(tmp_path))
+        monkeypatch.setattr(cli, 'dir', str(tmp_path))
```
manics: Both of these `monkeypatch` statements are needed, otherwise you either get a warning about an unset `OMERODIR` or `str(tmp_path)` is not used.

manics: The behaviour has changed again, now it works with only `monkeypatch.setenv('OMERODIR', str(tmp_path))`

joshmoore: That might mean that we should not use the environment variable anywhere other than in `cli.py` (cf. https://github.com/ome/omero-py/pull/14/files#diff-128953e35cc62025d88b5ef1c3eb882eR255) For that to work, though, we might have to default to `None` rather than `.dirname().dirname().dirname()` in https://github.com/ome/omero-py/blob/master/src/omero/cli.py#L103

joshmoore:
I assume that https://github.com/ome/omero-py/pull/160 et al may change the logic here, but better to get this test in earlier rather than later.

PR #143, opened 2019-12-04
Fix tables StringColumn unicode

manics:
Replacement for https://github.com/ome/omero-py/pull/133

See comment in hdfstorageV2.HdfStorage.append (https://github.com/ome/omero-py/commit/9e1d8e00873f76a2a2972a0c96de1494639c9c48) for the full explanation

Probably needs a doc update to say string column sizes must be given as the number of bytes not the number of unicode characters.

### DIFF
--- a/src/omero/columns.py
+++ b/src/omero/columns.py
@@ -21,11 +21,11 @@ from builtins import zip
 from builtins import range
 from builtins import object
 from future.utils import native, bytes_to_native_str, isbytes
-import sys
 import omero
 import Ice
 import IceImport
 IceImport.load("omero_Tables_ice")
+sys = __import__("sys")  # Python sys
 
 try:
     import numpy
--- a/src/omero/columns.py
+++ b/src/omero/columns.py
@@ -21,6 +21,7 @@ from builtins import zip
 from builtins import range
 from builtins import object
 from future.utils import native, bytes_to_native_str, isbytes
+import sys
 import omero
 import Ice
 import IceImport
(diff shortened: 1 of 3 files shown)


manics:
The last commit fails on Pytohn 2.7 ewiyth the inexplicable error
```
    def arrays(self):
        """
        Check for strings longer than the initialised column width
        """
        for v in self.values:
>           if sys.version_info >= (3, 0, 0):
E           AttributeError: 'module' object has no attribute 'version_info'
```
https://travis-ci.org/ome/omero-py/jobs/620734836

joshmoore:
There is also an `omero.sys` module which is likely what Py2 is picking up. There are existing import gymnastics to handle that.

joshmoore:
(Title is probably also a bit off now, eh? 🙂)

joshmoore:
Nice! A new and improved exception today:

```
ValidationException: exception ::omero::ValidationException {     serverStackTrace =      serverExceptionClass =      message = Maximum string (byte) length in column stringcol is 4 }
```

https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/262/testReport/OmeroPy.test.integration.tablestest.test_service/TestTables/testAllColumnsSameTable/

:+1:

manics:
It might be possible to simplify this by moving the conversion into `StringColumnI.arrays()` so bytes are always returned. This assumes `arrays()` isn't used anywhere else.

manics:
See last commit on https://github.com/ome/omero-py/pull/145/ re the last comment

PR #145, opened 2019-12-05
StringColumn: Handle all encoding inside class

manics:
Clean-up https://github.com/ome/omero-py/pull/143 by moving all changes into `StringColumnI`


src/omero/columns.py
```diff
@@ -25,7 +25,7 @@
 import Ice
 import IceImport
 IceImport.load("omero_Tables_ice")
-sys = __import__("sys")  # Python sys
+python_sys = __import__("sys")  # Python sys
```
manics: For some reason `sys = ` stopped working after some other change, hence `python_sys = `

PR #147, opened 2019-12-12
wheel

jburel:
add wheel to requirement
Following comment https://forum.image.sc/t/omero-python-3-5-6-milestone-for-upgrade-testing/32120/2
I had a closer look at the log of omero-install. I could see error like ```invalid command 'bdist_wheel'``` when installing omero-py (same for omero-web) on all OS

Check that you cannot see the error in the travis log with this PR

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -210,6 +210,7 @@ setup(
         'future',
         'numpy',
         'Pillow',
+        'wheel',
         'zeroc-ice>=3.6.4,<3.7',
     ],
     tests_require=[


dominikl:
Looks good. Can't see error in travis build and tried to pip install from this PR:
```
...
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/00/83/b4a77d044e78ad1a45610eb88f745be2fd2c6d658f9798a15e384b7d57c9/wheel-0.33.6-py2.py3-none-any.whl
...
```
👍

manics:
OMERO.py doesn't directly depend on wheel so I'm worried this is hiding some other problem. I think the original PR where it was Installed in omero-install is safer.

jburel:
The same will apply to omero-web.
I think is that omero-py and omero-web should be installed without "error" when user runs the ``pip`` command

jburel:
Discussed today. 
Plan is to review the dependencies during the bump 
We agree that ``wheel`` is not directly a dependency of omero-py.

PR #148, opened 2019-12-16
Ignore OMERO_HOME in bin/omero with warning

will-moore:
See https://forum.image.sc/t/omero-python-3-5-6-milestone-for-upgrade-testing/32120/3

OMERO_HOME shouldn't be used any more, but we still try to read it.
This now ignores OMERO_HOME (but prints a warning that it is being ignored).

To test:
 ```
$ export OMERO_HOME=/path/to/somewhere
# any omero command
$ omero -h
WARN: OMERO_HOME usage is deprecated in omero-py
```

### DIFF
--- a/bin/omero
+++ b/bin/omero
@@ -95,6 +95,7 @@ try:
     try:
         import omero.cli
     except ImportError as ie:
+        OMERODIR = os.environ.get('OMERODIR', None)
         print("*"*80)
         print("""
         ERROR: Could not import omero.cli! (%s)
@@ -115,9 +116,11 @@ try:
         --------------
         CWD=%s
         VERSION=%s
-        dist=%s
+        OMERO_EXE=%s
+        OMERODIR=%s
         PYTHONPATH=%s
-        """ % (os.getcwd(), sys.version.replace("\n"," "), top, sys.path))
+        """ % (os.getcwd(), sys.version.replace("\n", " "), top,
+               OMERODIR, sys.path))
         sys.exit(2)
 
     logging.basicConfig(level=logging.WARN)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: bin/omero [hunk 1]
```diff
@@ -122,9 +115,9 @@ try:
         --------------
         CWD=%s
         VERSION=%s
-        OMERO_HOME=%s
+        dist=%s
```
joshmoore: Perhaps OMERODIR= ?

will-moore: That might be confusing since this is not determined by the ``$OMERODIR`` env variable.
It's ``top = os.path.join(exe, os.pardir, os.pardir)``

joshmoore: Perhaps then OMERO_EXE and print OMERODIR separately?

bin/omero
```diff
@@ -59,18 +59,11 @@ def readlink(file=sys.argv[0]):
     file = os.path.abspath(file)
     return file
 
-omero_home = None
 if "OMERO_HOME" in os.environ:
-    omero_home = os.environ["OMERO_HOME"]
-    if not os.path.exists(omero_home):
-        print("OMERO_HOME=%s cannot be found" % omero_home)
-        sys.exit(3)
+    print("WARN: OMERO_HOME usage is deprecated in omero-py")
```
manics:
```suggestion
    print("WARN: OMERO_HOME usage is ignored in omero-py")
```
[Deprecated](https://en.wikipedia.org/wiki/Deprecation#Software_deprecation) implies it works but is not supported/recommended

joshmoore:
Same decision to be made on ignoring as in https://github.com/ome/omero-web/pull/98#issuecomment-566986551

PR #149, opened 2019-12-16
omero.client ensure args is not None

manics:
`c = omero.client('wss://idr.openmicroscopy.org/omero-ws')` should work instead of throwing an Exception

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -122,6 +122,8 @@ class BaseClient(object):
         # Reassigning based on argument type
 
         args, id, host, port, pmap = self._repair(args, id, host, port, pmap)
+        if not args:
+            args = []
 
         # hosturl overrides all other args
         hosturl = self._check_for_hosturl(host, port, pmap)
@@ -134,9 +136,7 @@ class BaseClient(object):
             args.append(self._get_endpoint_from_hosturl(hosturl))
 
         # Copying args since we don't really want them edited
-        if not args:
-            args = []
-        else:
+        if args:
             # See ticket:5516 To prevent issues on systems where the base
             # class of path.path is unicode, we will encode all unicode
             # strings here.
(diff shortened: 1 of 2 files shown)


jburel:
Without this PR:
``client = omero.client('ws://idr.openmicroscopy.org/omero-ws')`` or ``client = omero.client(host='ws://idr.openmicroscopy.org/omero-ws')`` will throw 
```
ModuleNotFoundError: No module named 'omero_ext.path'
```
With this PR:
``client = omero.client(host='ws://idr.openmicroscopy.org/omero-ws')`` or ``client = omero.client('ws://idr.openmicroscopy.org/omero-ws')`` do not throw.


This used to work on early dev version so this should be included in the coming 5.6 release
This prevents the usage of web socket (e.g. mybinder usage, demo scheduled 23/01)

joshmoore:
> ModuleNotFoundError: No module named 'omero_ext.path'

I'm unsure how this would be related. For me, this suggests that something is out of sync, since this has been fixed separately.

jburel:
Sorry I was not on master when I first run the command
The error without the PR is 
```
>>> client = omero.client(host='ws://idr.openmicroscopy.org/omero-ws')
  File "/Users/jmarie/Documents/git-repo/omero-py/target/omero/__init__.py", line 67, in client
    return omero.clients.BaseClient(*args, **kwargs)
  File "/Users/jmarie/Documents/git-repo/omero-py/target/omero/clients.py", line 134, in __init__
    args.append(self._get_endpoint_from_hosturl(hosturl))
AttributeError: 'NoneType' object has no attribute 'append'
```

jburel:
Tested in notebook too ``AttributeError: 'NoneType' object has no attribute 'append'
``
with dev9
Regression between dev4 and dev9

PR #157, opened 2020-01-06
Remove use and docs for bin/omero

joshmoore:
- testlib, etc. no longer search for dist/bin
 - CLI help text and generated docs (config_parser) omit the bin/
 - shellserver.py has been removed since it is unused


joshmoore:
Releasing as a quick dev10 before cutting final versions.

Issue #159, opened 2020-01-07
omero admin start doesn't warn about OMERODIR


joshmoore:
```
$ omero admin start
FATAL: OMERO directory does not exist: /opt/etc/grid
```

PR #160, opened 2020-01-07
prefs.py: config throw if OMERODIR not set

manics:
Partial fix for https://github.com/ome/omero-py/issues/159
This throws an exception. Ideally it'll be caught and wrapped into a nice error message in the CLI plugins but I this is at least better than nothing.

### DIFF
--- a/src/omero/plugins/prefs.py
+++ b/src/omero/plugins/prefs.py
@@ -66,6 +66,9 @@ def getprefs(args, dir):
 
 def _make_open_and_close_config(func, allow_readonly):
     def open_and_close_config(*args, **kwargs):
+        if not os.environ.get('OMERODIR'):
+            raise Exception('OMERODIR not set')
+
         args = list(args)
         self = args[0]
         argp = args[1]


joshmoore:
Looks like _not_ having it set is the case for a number of unit tests.

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#222](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/222/). See the [console output](https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/222/consoleText) for more details.
Possible conflicts:
  - PR #157 joshmoore 'Remove use and docs for bin/omero'
    - src/omero/testlib/__init__.py

--conflicts

joshmoore:
Not sure what would lead to it, but `str(self.dir)` would be `'None'` in `_cwd` with the new logic. I think https://github.com/joshmoore/omero-py/commit/abdc0be9d394414c289391aeb1343023bc587482 is correct.

manics:
If we don't mind a breaking change it might be easier to NOT pass `dir` down to each plugin, and instead rely on the plugin calling `get_omerodir()` if it really needs the dir?

joshmoore:
--exclude

```
19:21:59   File "/home/omero/workspace/OMERO-server/omero-virtualenv/lib/python3.6/site-packages/past/utils/__init__.py", line 95, in old_div
19:21:59     return a / b
19:21:59 TypeError: unsupported operand type(s) for /: 'str' and 'str'
19:21:59 Build step 'Execute shell' marked build as failure
```

https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-server/185/console

joshmoore:
Tried updating. Looks like it needs some test fixes as well:

```
    def _get_classpath_logback(self, args):
>       lib_client = self.ctx.dir / "lib" / "client"
E       TypeError: unsupported operand type(s) for /: 'NoneType' and 'str'
```

Issue #161, opened 2020-01-08
omero import needs to check for OMERODIR


joshmoore:
If OMERODIR is not set, the following error is printed.

```
09:17:12 + omero import -d Dataset:4851 test.fake --debug ERROR
09:17:13 Cannot get JAR files from '/home/omero/workspace/OMERO-training/omero-virtualenv/lib/lib/client' (No such file or directory)
```

Easy fix: check before running and show the user a helpful message like "Please set OMERODIR".

Harder fix: Automatically download the importer jars.

Workaround: Set OMERODIR :smile:

joshmoore:
The above comes from the OMERO-training job. I renamed the `OMERO_SERVER_DIST` bash variable to be `OMERODIR` correcting the issue.

glyg:
Hi Josh, I have the same error (with the `lib/lib/client` path) when trying to import a directory from a remote client. Is there a quick fix (neither `OMERO_SERVER_DIST` nor `OMERODIR` are set in my environment)? 
omero is version 5.6.2

G.

glyg:
sorry I just need to define OMERODIR - sorry for the noise everything is ok

PR #162, opened 2020-01-08
import --fetch-jars

manics:
Thought I'd see how difficult it is to automatically fetch the jars....

```
$ rm -rf ~/omero/jars/

$ omero import test.fake
Downloading https://downloads.openmicroscopy.org/omero/latest/OMERO.java.zip

... import as normal...
```

Closes https://github.com/ome/omero-py/pull/161

### DIFF
--- a/src/omero/plugins/import.py
+++ b/src/omero/plugins/import.py
@@ -32,11 +32,23 @@ from past.builtins import basestring
 from builtins import object
 import os
 import csv
+from shutil import copyfileobj
 import sys
 import shlex
 
+try:
+    from io import BytesIO
+    from urllib.request import urlopen
+except ImportError:
+    # Python 2
+    from StringIO import StringIO as BytesIO
+    from urllib import urlopen
+from zipfile import ZipFile
+
+
 from omero.cli import BaseControl, CLI
 import omero.java
+from omero.util import get_omero_userdir
 from omero_ext.argparse import SUPPRESS
 from omero_ext.path import path
 
@@ -100,6 +112,11 @@ OUTPUT_CHOICES = ["ids", "legacy", "yaml"]
 SKIP_CHOICES = ['all', 'checksum', 'minmax', 'thumbnails', 'upgrade']
 NO_ARG = object()
 
+OMERO_JAVA_ZIP = (
+    'https://downloads.openmicroscopy.org/omero/5.6.0-m5/artifacts/'
+    'OMERO.java-5.6.0-m5-ice36-b129.zip'
+)
+
 
 class CommandArguments(object):
 
@@ -118,6 +135,7 @@ class CommandArguments(object):
             "port", "password", "group", "create", "func",
             "bulk", "prog", "user", "key", "path", "logprefix",
             "JAVA_DEBUG", "quiet", "server", "depth", "clientdir",
+            "fetch_jars",
             "sudo")
         self.set_login_arguments(ctx, args)
         self.set_skip_arguments(args)
@@ -315,6 +333,9 @@ class ImportControl(BaseControl):
             "--logback", type=str,
             help="Path to a logback xml file. "
             " Default: etc/logback-cli.xml")
+        add_python_argument(
+            "--fetch-jars", action="store_true",
+            help="Download OMERO.java jars")
 
         # The following arguments are strictly passed to Java
         name_group = parser.add_argument_group(
@@ -486,11 +507,17 @@ class ImportControl(BaseControl):
         parser.set_defaults(func=self.importer)
 
     def importer(self, args):
+        if args.fetch_jars:
+            self.download_omero_java()
+            return
 
         if args.clientdir:
-            client_dir = path(args.clientdir)
+            client_dirs = [path(args.clientdir)]
         else:
-            client_dir = self.ctx.dir / "lib" / "client"
+            client_dirs = [
+                get_omero_userdir() / "jars",
+                self.ctx.dir / "lib" / "client",
+            ]
         etc_dir = old_div(self.ctx.dir, "etc")
         if args.logback:
             xml_file = path(args.logback)
@@ -498,13 +525,17 @@ class ImportControl(BaseControl):
             xml_file = old_div(etc_dir, "logback-cli.xml")
         logback = "-Dlogback.configurationFile=%s" % xml_file
 
-        try:
-            classpath = [file.abspath() for file in client_dir.files("*.jar")]
-        except OSError as e:
-            self.ctx.die(102, "Cannot get JAR files from '%s' (%s)"
-                         % (client_dir, e.strerror))
+        classpath = [
+            file.abspath()
+            for client_dir in client_dirs
+            if client_dir.exists()
+            for file in client_dir.files("*.jar")
+        ]
         if not classpath:
-            self.ctx.die(103, "No JAR files found under '%s'" % client_dir)
+            self.ctx.die(103, (
+                "No JAR files found under '%s'.\n"
+                "Run 'omero import --fetch-jars' and re-try your import" %
+                ','.join(client_dirs)))
 
         command_args = CommandArguments(self.ctx, args)
         xargs = [logback, "-Xmx1024M", "-cp", os.pathsep.join(classpath)]
@@ -516,6 +547,23 @@ class ImportControl(BaseControl):
             self.bulk_import(command_args, xargs)
         else:
             self.do_import(command_args, xargs)
+
+    def download_omero_java(self):
+        print("Downloading %s ..." % OMERO_JAVA_ZIP, file=sys.stderr)
+        jars_dir = get_omero_userdir() / "jars"
+        if not jars_dir.exists():
+            jars_dir.mkdir()
+        resp = urlopen(OMERO_JAVA_ZIP)
+        content = resp.read()
+        content = BytesIO(content)
+        zipfile = ZipFile(content)
+        for f in zipfile.namelist():
+            if f.endswith('.jar'):
+                jar = (jars_dir / os.path.basename(f))
+                source = zipfile.open(f)
+                target = jar.open('wb')
+                with source, target:
+                    copyfileobj(source, target)
 
     def do_import(self, command_args, xargs, mode="w"):
         out = err = None
--- a/src/omero/plugins/import.py
+++ b/src/omero/plugins/import.py
@@ -106,7 +106,7 @@ SKIP_CHOICES = ['all', 'checksum', 'minmax', 'thumbnails', 'upgrade']
 NO_ARG = object()
 
 OMERO_JAVA_ZIP = (
-    'https://downloads.openmicroscopy.org/omero/latest/OMERO.java.zip'
+    'https://downloads.openmicroscopy.org/omero/{version}/OMERO.java.zip'
 )
 
 
@@ -326,8 +326,8 @@ class ImportControl(BaseControl):
             help="Path to a logback xml file. "
             " Default: etc/logback-cli.xml")
         add_python_argument(
-            "--fetch-jars", action="store_true",
-            help="Download OMERO.java jars")
+            "--fetch-jars", type=str,
+            help="Download this version of OMERO.java jars")
 
         # The following arguments are strictly passed to Java
         name_group = parser.add_argument_group(
@@ -502,7 +502,7 @@ class ImportControl(BaseControl):
         if args.fetch_jars:
             if args.path:
                 self.ctx.err('WARNING: Ignoring extra arguments')
-            self.download_omero_java()
+            self.download_omero_java(args.fetch_jars)
             return
 
         if args.clientdir:
@@ -555,12 +555,13 @@ class ImportControl(BaseControl):
         else:
             return None, omero_java_txt
 
-    def download_omero_java(self):
-        self.ctx.err("Downloading %s" % OMERO_JAVA_ZIP)
+    def download_omero_java(self, version):
+        omero_java_zip = OMERO_JAVA_ZIP.format(version=version)
+        self.ctx.err("Downloading %s" % omero_java_zip)
         jars_dir, omero_java_txt = self._userdir_jars(parentonly=True)
         if not jars_dir.exists():
             jars_dir.mkdir()
-        resp = urlopen(OMERO_JAVA_ZIP)
+        resp = urlopen(omero_java_zip)
         zipfile = ZipFile(BytesIO(resp.read()))
         topdirs = set(f.filename.split(
             os.path.sep)[0] for f in zipfile.filelist if f.is_dir())
--- a/src/omero/plugins/import.py
+++ b/src/omero/plugins/import.py
@@ -498,6 +498,29 @@ class ImportControl(BaseControl):
 
         parser.set_defaults(func=self.importer)
 
+    def _get_classpath_logback(self, args):
+        if args.clientdir:
+            client_dirs = [path(args.clientdir)]
+        else:
+            client_dirs = [self.ctx.dir / "lib" / "client"]
+            omero_java_dir, _ = self._userdir_jars()
+            if omero_java_dir:
+                client_dirs.append(omero_java_dir)
+        etc_dir = old_div(self.ctx.dir, "etc")
+        if args.logback:
+            xml_file = path(args.logback)
+        else:
+            xml_file = old_div(etc_dir, "logback-cli.xml")
+        logback = "-Dlogback.configurationFile=%s" % xml_file
+
+        classpath = [
+            file.abspath()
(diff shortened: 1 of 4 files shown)


manics:
--exclude

For `5.6.1`

joshmoore:
Two pretty self-unsimilar options come to mind:
 - a more maven-like strategy (see [jgo](https://pypi.org/project/jgo/))
 - doing this without a flag (but perhaps a warning) to a temporary location; a flag could make it permanent: `import --fetch-jars=$LOCATION`

manics:
OMERO.java is a 125 MB download, I don't think we shold do that automatically

joshmoore:
I'd bet most users would prefer for us to just do it, though maybe not repeatedly.

manics:
Hard-coding the OMERO.java version could potentially be avoided by fetching the server version:
```
        c = self.ctx.conn(args)
        version = c.sf.getConfigService().getVersion()
```
Though this assumes `OMERO.java version == OMERO.server version`. It also requires the logic from https://github.com/ome/omego/blob/v0.7.0/omego/artifacts.py#L391 to figure out the build number.

joshmoore:
cc: @sbesson 

I'm still for having this somewhat transport rather than a solely being a new argument that needs calling.

manics:
To make this more maintainable we could:
- Move the [omego artifacts plugin](https://github.com/ome/omego/blob/master/omego/artifacts.py) to a new omero cli plugin e.g. `package-download`
- Use that plugin/library to download the jars for the latest version (or perhaps the latest major version matchinf the OMERO-py version?)

manics:
As requested I've made it automatically download the jars if required, the latest https://downloads.openmicroscopy.org/omero/latest/OMERO.java.zip will be fetched.

If you want to force a re-download, update or download a different version of the jars pass a version to the `--fetch-jars` argument:

```
$ rm -rf ~/omero/jars
```
```
$ omero import test.fake
Downloading https://downloads.openmicroscopy.org/omero/latest/OMERO.java.zip

<import as normal>

$ cat ~/omero/jars/OMERO.java
OMERO.java-5.6.1-ice36-b225/ OMERO.java.txt
```

```
$ omero import --fetch-jars 5.5
Downloading https://downloads.openmicroscopy.org/omero/5.5/OMERO.java.zip

$ cat ~/omero/jars/OMERO.java.txt
OMERO.java-5.5.1-ice36-b122
```

```
$ omero import --fetch-jars latest
Downloading https://downloads.openmicroscopy.org/omero/latest/OMERO.java.zip

$ cat ~/omero/jars/OMERO.java.txt
OMERO.java-5.6.1-ice36-b225
```

The available versions strings are whatever `https://downloads.openmicroscopy.org/omero/{version}/OMERO.java.zip` redirects are in place on our downlaods server.

manics:
(Forgot to push the last commit before commenting)

src/omero/plugins/import.py
```diff
@@ -517,6 +542,43 @@ def importer(self, args):
...
+        omero_java_zip = OMERO_JAVA_ZIP.format(version=version)
+        self.ctx.err("Downloading %s" % omero_java_zip)
+        jars_dir, omero_java_txt = self._userdir_jars(parentonly=True)
+        if not jars_dir.exists():
+            jars_dir.mkdir()
+        resp = urlopen(omero_java_zip)
+        zipfile = ZipFile(BytesIO(resp.read()))
+        topdirs = set(f.filename.split(
+            os.path.sep)[0] for f in zipfile.filelist if f.is_dir())
+        if len(topdirs) != 1:
```
jburel, line 568: zipfile will not be closed if it failed here. ``try/finally`` will be more suitable

src/omero/plugins/import.py
```diff
@@ -517,6 +542,43 @@ def importer(self, args):
...
+            self.ctx.die(
+                108, 'Expected one top directory in OMERO.java.zip: {}'.format(
+                    topdirs))
+        topdir = topdirs.pop()
+        if os.path.isabs(topdir):
+            self.ctx.die(
+                108, 'Unexpected absolute paths in OMERO.java.zip: {}'.format(
+                    topdir))
+        zipfile.extractall(jars_dir)
+        zipfile.close()
```
jburel, line 578: see comment above

src/omero/plugins/import.py
```diff
@@ -517,6 +542,43 @@ def importer(self, args):
...
+        omero_java_txt.write_text(topdir)
+        resp.close()
```
jburel, line 580: same

src/omero/plugins/import.py
```diff
@@ -486,25 +499,37 @@ def add_advjava_argument(*args, **kwargs):
...
-        logback = "-Dlogback.configurationFile=%s" % xml_file
-
-        try:
-            classpath = [file.abspath() for file in client_dir.files("*.jar")]
-        except OSError as e:
-            self.ctx.die(102, "Cannot get JAR files from '%s' (%s)"
-                         % (client_dir, e.strerror))
-        if not classpath:
-            self.ctx.die(103, "No JAR files found under '%s'" % client_dir)
+        for attempt in [0, 1]:
```
joshmoore, line 508: The `attempt` logic seems more complicated than it needs to be. Would refactoring out the classpath logic help pre-check?

src/omero/plugins/import.py
```diff
@@ -517,6 +542,43 @@ def importer(self, args):
         else:
             self.do_import(command_args, xargs)
 
+    def _userdir_jars(self, parentonly=False):
+        user_jars = get_omero_userdir() / 'jars'
```
joshmoore, line 546: No objections to this unless we foresee moving away from `~/omero` soonish. If so, then perhaps something like [user_cache_dir()/omero](https://pypi.org/project/appdirs/) would be safer (so people don't have to re-download).

src/omero/plugins/import.py
manics: Thanks for spotting this, I've switched to a context manager

src/omero/plugins/import.py
manics: Done!

src/omero/plugins/import.py
manics: It's a fair point, but I think having two different dirs for storing user state is worse than re-downloading. If there's a concrete suggestion for the location of `user_cache_dir()` shouldn't we modify `get_omero_userdir()` instead?

manics:
I had to mess around with the Ice build due to openssl issues. What I think is happening is that https://github.com/ome/zeroc-ice-py-manylinux/releases/download/0.2.0/zeroc_ice-3.6.5-cp36-cp36m-manylinux2010_x86_64.whl included a bundled openssl library which is somehow incompatible with Travis python.

https://github.com/ome/zeroc-ice-ubuntu1804/releases/download/0.3.0/zeroc_ice-3.6.5-cp36-cp36m-linux_x86_64.whl doesn't work, presumably because Travis Pytohn 3.6 is different from the Ubuntu syste, Python 3.6.

A possible way around this is to build wheels specific to Travis, but for now I've changed this to use the default source package.

Note the reason https requests are being made is that although most tests in https://github.com/ome/omero-py/blob/master/test/unit/clitest/test_import.py that require the jars are disabled with `@pytest.mark.skipif(OMERODIR is False, reason="Needs client dir")` some of them aren't marked:
https://github.com/ome/omero-py/blob/41f82a84a88a3a9153a2c48702f1e4c28fa86167/test/unit/clitest/test_import.py#L38
and are instead allowed to return a non-zero exit code if the jars are missing:
https://github.com/ome/omero-py/blob/41f82a84a88a3a9153a2c48702f1e4c28fa86167/test/unit/clitest/test_import.py#L134-L138
```
$ omero import --advanced-help
Error: Could not find or load main class ome.formats.importer.cli.CommandLineImporter
```

With this PR the jars are now automatically downloaded.

If you're wondering whether the other import tests marked `@pytest.mark.skipif(OMERODIR is False, reason="Needs client dir")` should now pass, some of them do, but some of them are broken, suggesting we never run these unit tests with `OMERODIR` defined.

manics:
See https://github.com/ome/omero-py/pull/231 (on top of this one) for the fixed import tests. I can merge that PR into this one if you think it's appropriate.

src/omero/plugins/import.py
joshmoore: My proposal for moving forward would be to place an `omero` subdirectories into existing directories like `.config`, `.cache`. If that's controversial at all, let's not block this PR.

joshmoore:
> I can merge that PR into this one if you think it's appropriate.

:+1:

Only other comment at this stage is whether or not `omero import` could print out the used version.

manics:
I've:
- rebased
- added the 4 commits from https://github.com/ome/omero-py/pull/231
- added `omero.util.get_omero_user_cache_dir`. Currently this is the same as `get_omero_userdir` but will change with https://github.com/ome/omero-py/issues/210. Since it's quite a big change I don't think we should do that in this PR.

src/omero/plugins/import.py
```diff
@@ -517,6 +544,43 @@ def importer(self, args):
...
+            return None, omero_java_txt
+
+    def download_omero_java(self, version):
+        omero_java_zip = OMERO_JAVA_ZIP.format(version=version)
+        self.ctx.err("Downloading %s" % omero_java_zip)
+        jars_dir, omero_java_txt = self._userdir_jars(parentonly=True)
+        if not jars_dir.exists():
+            jars_dir.mkdir()
+        with urlopen(omero_java_zip) as resp:
+            with ZipFile(BytesIO(resp.read())) as zipfile:
```
joshmoore, line 567: It may fall more under https://github.com/ome/omero-py/issues/217, but some feedback of the download progress at least in an interactive terminal will probably prevent consternation.

src/omero/plugins/import.py
```diff
@@ -485,26 +498,40 @@ def add_advjava_argument(*args, **kwargs):
...
-    def importer(self, args):
-
+    def _get_classpath_logback(self, args):
         if args.clientdir:
-            client_dir = path(args.clientdir)
+            client_dirs = [path(args.clientdir)]
         else:
-            client_dir = self.ctx.dir / "lib" / "client"
+            client_dirs = [self.ctx.dir / "lib" / "client"]
+            omero_java_dir, _ = self._userdir_jars()
```
joshmoore, line 506: Could this be done _only if_ `lib/client` doesn't exist which would give you a chance to print, "Using downloading version: $VERSION"

src/omero/plugins/import.py
manics: I could look into TQDM https://github.com/tqdm/tqdm

But that's another dependency- should split omero-cli out of omero-py to avoid installing UI dependencies before going further?

src/omero/plugins/import.py
manics: Done, jars are only automatically downloaded if both `--clientdir` argument and `lib/client` are absent, otherwise it reverts to the original failure message

src/omero/plugins/import.py
joshmoore: Yeah, I imagine that's one we put on the backburner.

src/omero/plugins/import.py
joshmoore:
```
(demo) /tmp $omero import a.fake -- --debug=ERROR
Using OMERO.java-5.6.1-ice36-b225
Previously logged in to localhost:4064 as root

(demo) /tmp $omero import --clientdir=/opt/ome9/dist/lib/client a.fake -- --debug=ERROR
Previously logged in to localhost:4064 as root

(demo) /opt/ome9/dist $omero import --clientdir=/opt/ome9/dist/lib/client a.fake -- --debug=ERROR
Previously logged in to localhost:4064 as root
```

:+1:

manics:
`logback-cli.xml`: I think it would make sense for it to be part of the OMERO.java zip, I don't know whether it's better to be inside a jar or outside. Since this PR now distinguishes between the downloaded OMERO.java and the default `lib/client` it's should be easy to adjust the logback path.

joshmoore:
Ok. So if we update https://github.com/ome/openmicroscopy/blob/develop/build.xml#L302 we'll get logback-cli.xml into that directory and things should just magically start working... (Worth getting that location set now?)

Otherwise, :+1:

manics:
After inserting an extra `print` to check:
```diff
$ git diff
diff --git a/src/omero/plugins/import.py b/src/omero/plugins/import.py
index d23312c1..c890f928 100644
--- a/src/omero/plugins/import.py
+++ b/src/omero/plugins/import.py
@@ -530,6 +530,7 @@ class ImportControl(BaseControl):
                     103, "No JAR files found under '%s'" % client_dir)

         logback = "-Dlogback.configurationFile=%s" % xml_file
+        print(logback)
         return classpath, logback

     def importer(self, args):
```
```
$ OMERO_USERDIR=/tmp/xx omero import --advanced-help
Using OMERO.java-5.6.1-ice36-b225
-Dlogback.configurationFile=/tmp/xx/jars/OMERO.java-5.6.1-ice36-b225/libs/logback-cli.xml
...
```
```
$ OMERODIR=/openmicroscopy/dist/ omero import --advanced-help
-Dlogback.configurationFile=/openmicroscopy/dist/etc/logback-cli.xml
...
```

src/omero/plugins/import.py
```diff
@@ -485,26 +498,51 @@ def add_advjava_argument(*args, **kwargs):
...
             xml_file = old_div(etc_dir, "logback-cli.xml")
+
+        classpath = []
+        if client_dir and client_dir.exists():
+            classpath = [f.abspath() for f in client_dir.files("*.jar")]
+        if auto_download:
+            if classpath:
+                self.ctx.err('Using {}'.format(omero_java_txt.text()))
+                if not args.logback:
+                    xml_file = client_dir / "logback-cli.xml"
```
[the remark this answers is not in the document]
manics: Yes, but since this is a new feature I think it's fine, especially as it sounds like we have a pending release of OMERO.java in a few weeks and the default is to download `latest`

src/omero/plugins/import.py
```diff
@@ -35,8 +35,19 @@
 import sys
 import shlex
 
+try:
+    from io import BytesIO
+    from urllib.request import urlopen
+except ImportError:
+    # Python 2
```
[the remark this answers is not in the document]
manics: Are you looking at an old commit?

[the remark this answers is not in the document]
sbesson: I have no idea how this ended up there...

joshmoore:
Thanks again, @manics. I think this is going to make things much nicer for people since the move to `OMERODIR`. Merging.

(I'd add that as it gets usage there could well be feature requests for the likes of `--list-jars`, `--del-jars=$OLD_VERSION`, etc. which may even be best-placed in its own plugin.)

sbesson:
Should we release the current state of `master` as `omero-py 5.8.0` or would you like to include #233 ?

manics:
It'd be nice to include #233, but that may need futher discussions, so I wouldn't say it's a blocker. Feel free to push to it if you want.

## PRs merged since 5.7.1:
Using https://gist.github.com/manics/383b4481e1c3a193d224016b01a5d52a
`GH_TOKEN=$(git config github.token) GH_REPO=omero-py ../generate_release_notes/generate_release_notes.py v5.7.1 master --version v5.8.0`

- import --fetch-jars ([#162](https://github.com/ome/omero-py/pull/162))
- fix logging attribute error ([#226](https://github.com/ome/omero-py/pull/226))
- Add PyYAML as a dependency of omero-py ([#228](https://github.com/ome/omero-py/pull/228))
- bin/omero entrypoint ([#229](https://github.com/ome/omero-py/pull/229))
- hql.py: when querying masks, filter `bytes` field ([#230](https://github.com/ome/omero-py/pull/230))

https://github.com/ome/omero-py/pull/226 is the only bug fix so if that's not urgent I think we can wait a bit.

src/omero/plugins/import.py
```diff
@@ -517,6 +555,43 @@ def importer(self, args):
...
+            return user_jars, omero_java_txt
+        omero_java_dir = None
+        if omero_java_txt.exists():
+            omero_java_dir = omero_java_txt.text().strip()
+            return user_jars / omero_java_dir / 'libs', omero_java_txt
+        else:
+            return None, omero_java_txt
+
+    def download_omero_java(self, version):
+        omero_java_zip = OMERO_JAVA_ZIP.format(version=version)
```
joshmoore, line 572: It occurs to me for the testing of https://github.com/ome/openmicroscopy/pull/6255 that `version` should likely have been used directly if it started with `http(s)://`

joshmoore: see: https://github.com/ome/omero-py/pull/303

PR #169, opened 2020-01-09
5.6 cleanup

joshmoore:
Few final cleanups in preparation for 5.6.0:

 - [x] drop `omero_setup.py`. see https://github.com/ome/openmicroscopy/pull/6201
 - [x] drop uses of `lib/python` and `lib/fallback`


joshmoore:
No final objections. Beginning the non-dev release of the Python modules in preparation for OMERO.server 5.6.0.

PR #174, opened 2020-01-27
Add show-uuid option to sessions who

chris-allan:
Adds a `--show-uuid` option to sessions who

### DIFF
--- a/src/omero/plugins/sessions.py
+++ b/src/omero/plugins/sessions.py
@@ -227,8 +227,12 @@ class SessionsControl(UserGroupControl):
             "--no-purge", dest="purge", action="store_false",
             help="Do not remove inactive sessions")
 
-        parser.add(sub, self.who, (
+        who = parser.add(sub, self.who, (
             "List all active server sessions\n\n" + WHOHELP))
+        who.add_argument(
+            "--show-uuid", help="Show uuids for sessions",
+            action="store_true")
+
 
         keepalive = parser.add(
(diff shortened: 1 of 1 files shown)


jburel:
I noticed that the info was available in the deprecated method ``omero sessions list`` but not in the recommended method ``omero sessions who`` when reviewing some of @mtbc work.
We discussed adding it later on. 
Thanks

chris-allan:
Thanks, @jburel.

`sessions list` applies to sessions visible to the OMERO CLI whereas `sessions who` is querying the server for all the active sessions. Not sure what you mean by deprecated.

jburel:
apologies I meant ``omero admin sessionlist``

chris-allan:
Anything else from `omero admin sessionlist` we want to bring over?

/cc @joshmoore, @mtbc, @manics

jburel:
Looking at the field: the  ``id`` was the only one missing 
```node | session                              | started                  | owner  | agent          | notes   ```
notes indicate if it the current session but this is now indicated using ``*``

chris-allan:
I guess just `node`? Not sure that's important information at all. Maybe in a case where there are multiple nodes off the same database? Seems like a big edge case that while theoretically possible is almost never a reality.

chris-allan:
@jburel: Can we get this merged or is there something else we need to do?

jburel:
Thanks for the reminder. I forgot about it while on the road. 
Merging so it can be included in the coming release. @manics will re-run the script for changelog

PR #176, opened 2020-02-06
Extend output of 'omero version'

joshmoore:
Intent is to clarify that different version numbers that a user may need
to be aware of. `omero_version` remains the main output (on stdout) but
reference is made on stderr to "OMERO.py" and if OMERODIR is accessible,
then a version is also printed for OMERO.server.

### DIFF
--- a/src/omero/plugins/basics.py
+++ b/src/omero/plugins/basics.py
@@ -28,6 +28,7 @@ from collections import defaultdict
 from omero.cli import BaseControl
 from omero.cli import CLI
 from omero.cli import VERSION
+from omero.cli import OMERODIR
 
 
 class QuitControl(BaseControl):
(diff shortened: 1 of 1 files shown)


joshmoore:
e.g.

```
$ omero version
OMERO.py version:
5.6.1.dev1
OMERO.server version:
5.6.1-ice36-SNAPSHOT
```

manics:
Probably fine for interactive users. Not so great if this is intended to be parsed. If this goes in now are further changes possible or would that be considered breaking?

joshmoore:
I'd think for the parsing-user, we'd introduce a new flag and/or command, e.g. `omero version --json`, since I don't think there's a way that we can make this parseable without already being breaking.

joshmoore:
Cheers. Merging. Leaving someone else to drive a release if it's wanted/needed.

PR #180, opened 2020-02-10
download: Check for null file size

joshmoore:
If an unexpected download ID is given, the client can fail with:

```
$ omero download 123456 my.ome.tif
...
  File "OMERO/bin/omero", line 130, in <module>
    rv = omero.cli.argv()
  File "OMERO/lib/python/omero/cli.py", line 1620, in argv
    cli.invoke(args[1:])
  File "OMERO/lib/python/omero/cli.py", line 1095, in invoke
  File "OMERO/lib/python/omero/cli.py", line 1172, in onecmd
  File "OMERO/lib/python/omero/cli.py", line 1254, in execute
    args.func(args)
  File "OMERO/lib/python/omero/plugins/download.py", line 66, in __call__
    client.download(orig_file, target_file)
  File "OMERO/lib/python/omero/clients.py", line 863, in download
    if block_size > ofile.size.val:
```

(line numbers are from 5.4.10)

cc: @chris-allan

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -944,7 +944,14 @@ class BaseClient(object):
             ofile = self.__sf.getQueryService().get(
                 "OriginalFile", ofile.id.val, ctx)
 
-            if ofile.size is not None and block_size > ofile.size.val:
+            if ofile.size is None:
+                name = omero.rtypes.unwrap(ofile.name)
+                mimetype = omero.rtypes.unwrap(ofile.mimetype)
+                raise omero.ClientError(
+                    ("invalid size for OriginalFile '%s' "
+                     "(mimetype:%s)") % (name, mimetype))
+
+            if block_size > ofile.size.val:
                 block_size = ofile.size.val
 
             prx.setFileId(ofile.id.val, ctx)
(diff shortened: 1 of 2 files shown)


chris-allan:
Thanks, @joshmoore. @kkoz is working on a variety of scenarios surrounding this. We were unable to find anything other than broken imports where `size` was `null` but there were also lots of other conditions that we found were the error checking was sub par. I'll let him comment when he's ready.

joshmoore:
:+1: Also feel free to just open separately and I can close this. Right now I'm looking into what the behavior is when one accidentally tries to download a `Directory` or a `Repository`. If you know what the original target was (called `123456` above) that would be useful.

chris-allan:
Yep, those are exactly conditions we were seeing as well. Also seen with trying to download an OMERO.tables instance. The original `123456` target was what we could only assume to be a broken import from many years ago. File was not actually in the `ManagedRepository` and the OriginalFile instance only partially filled out. The reason it was selected was misunderstanding the documentation and assuming `123456` was an Image ID (requiring `Image:123456` as a command line option) rather than an OriginalFile ID.

joshmoore:
Minor update to hide the stack trace from the user, e.g.:

```
ClientError: invalid size for OriginalFile 'Meas_04(2011-02-22_19-24-25)' (mimetype:Directory)
```

joshmoore:
--exclude

PR #181, opened 2020-02-11
Get file size for download from the RawFileStore instead of the DB

kkoz:
Similar to https://github.com/ome/omero-py/pull/180
This PR is to fix the issue where attempting to download a file with a null size in `originalfile` table results in the following:
```Traceback (most recent call last):
  File "OMERO/bin/omero", line 130, in <module>
    rv = omero.cli.argv()
  File "OMERO/lib/python/omero/cli.py", line 1620, in argv
    cli.invoke(args[1:])
  File "OMERO/lib/python/omero/cli.py", line 1095, in invoke
  File "OMERO/lib/python/omero/cli.py", line 1172, in onecmd
  File "OMERO/lib/python/omero/cli.py", line 1254, in execute
    args.func(args)
  File "OMERO/lib/python/omero/plugins/download.py", line 66, in __call__
    client.download(orig_file, target_file)
  File "OMERO/lib/python/omero/clients.py", line 863, in download
    if block_size > ofile.size.val:
```
If, however, we retrieve the size from the RawFileStore instead of the DB record, we can get a size for the file even if the DB field is null.

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -944,13 +944,20 @@ class BaseClient(object):
             ofile = self.__sf.getQueryService().get(
                 "OriginalFile", ofile.id.val, ctx)
 
-            if block_size > ofile.size.val:
-                block_size = ofile.size.val
-
             prx.setFileId(ofile.id.val, ctx)
+            size = None
+            if prx.size() is None:
+                name = omero.rtypes.unwrap(ofile.name)
+                mimetype = omero.rtypes.unwrap(ofile.mimetype)
+                raise omero.ClientError(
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks, @kkoz. You might want to also copy over the `ClientError` handling from #180.

jburel:
Wrote a test (I will open a PR against openmicroscopy)
Without this PR:
```if block_size > ofile.size.val:
E           AttributeError: 'NoneType' object has no attribute 'val'
```
With this PR
the test is green

jburel:
Client side handling has been copied over
Merging

PR #182, opened 2020-02-12
Update changelog and version for 5.6.1

manics:
Changelog notes mostly generated from https://gist.github.com/manics/383b4481e1c3a193d224016b01a5d52a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 5.6.1 (February 2020)
+
+- travis jobs: enable Windows ([#134](https://github.com/ome/omero-py/pull/134))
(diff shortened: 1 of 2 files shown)


jburel:
Side note: The gist will be useful to have in a central place considering the number of repository we have to handle,
This also means that as a team we need to work on the PR title a bit more

manics:
I assume @joshmoore has something that he used for the last update

joshmoore:
I have previous just used `git log --oneline --first-parent`. @chris-allan has suggested using github_changelog_generator. Regardless of which technique, I'm very much for our PR titles being end-user consumable.

jburel:
Changes look good
Please do go ahead.

jburel:
Thanks @manics

jburel:
Travis is green. Merging and tagging

Issue #2, opened 2019-08-06
Use black with precommit for automatic formatting


manics:
Get rid of annoying travis failures due to forgetting to run flake8 by making everything automatic:
- https://black.readthedocs.io/en/stable/index.html
- https://black.readthedocs.io/en/stable/version_control_integration.html

sbesson:
Given the impact of reformatting the whole codebase, this is currently not in scope or a priority.

PR #184, opened 2020-02-18
Auto-format code with black

manics:
Closes https://github.com/ome/omero-py/issues/2

Opening for discussion.
- First commit configures pre-commit to run the [`black` autoformatter](https://black.readthedocs.io/en/stable/)
- Second commit is the result of `pre-commit run --all-files`

In future code can be formatted by running `pre-commit run --all-files` locally.

--exclude
Since this will otherwise conflict with everything!


sbesson:
Closing for now as this is conflicting and will need additional work/considerations if we want to introduce black

PR #185, opened 2020-02-19
config: Escape quotes in `Default:` section

joshmoore:
At some point, sphinx started converting quotes to smartquotes
by default. This attempts to escape them only as necessary where
users will copy and paste values. It may be necessary, however,
to disable them globally.

see:
 - https://forum.image.sc/t/omero-web-installation-error-on-version-5-6-0/34150/3
 - http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-smartquotes

### DIFF
--- a/src/omero/install/config_parser.py
+++ b/src/omero/install/config_parser.py
@@ -421,6 +421,7 @@ class PropertyParser(object):
                 if not p.val:
                     v = "[empty]"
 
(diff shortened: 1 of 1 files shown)


mtbc:
Also see https://trello.com/c/OPOGgvps/503-in-options-renders-badly, this might help.

joshmoore:
Cheers.

PR #190, opened 2020-03-03
List preferred conda installation first

dominikl:
If conda is the preferred way to install, it should be mentioned first.

See also https://github.com/ome/ome-documentation/pull/2104


jburel:
The order makes sense

PR #194, opened 2020-03-05
Support Python 3 asyncio concurrency

manics:
Adds support for creating an OMERO session and Ice services that can be used with [`asyncio`](https://docs.python.org/3.6/library/asyncio.html).

- https://github.com/ome/omero-py/commit/00707b7df02f062ceba9328a688682d3acf25b52 creates an exact copy of `createSession` called `createAsyncSession`
- https://github.com/ome/omero-py/commit/a6583d0b80aeb38148234f628d5a741af63a8577 modifies `createAsyncSession` to create an async session

Example:
https://gist.github.com/manics/22583d9404022f1b57df0e40f97b4cd3

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -33,6 +33,7 @@ import Ice
 import re
 import ssl
 import uuid
+from .iceasync import AsyncService, AsyncSession
 
 from past.builtins import basestring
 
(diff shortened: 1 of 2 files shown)


manics:
Does anyone have opinions on how to support this usage? I initially implemented it as a separate python module, but the async version of `createSession`, `createAsyncSession`, is problematic to implement outside this module. `createSession` uses __private variables, so re-implementing this method in another class requires one of:
- break the encapsulation and unmangle the private variables: e.g. `self.__sf` is referenced as `self._BaseClient__sf`
- duplicate all private variables and methods called by `createSession`: `createAsyncSession` has to call other class methods that also reference some __private variables, so if those private variables are duplicated in the subclass then all methods that reference them must also be duplicated
- refactor `createSession` to support overriding: potentially dangerous given the complexity of the existing code

joshmoore:
I'm for (3) refactoring. Happy to help.

sbesson:
Closing as there has been no activity  for the last 4-5 years here.

PR #197, opened 2020-03-10
processor: set OMERODIR

joshmoore:
https://github.com/ome/openmicroscopy/pull/6222 is currently
failing since the script run by the test is not able to find
the importer jars. With OMERODIR set in the processor server
the jars should be detected.


dominikl:
To me too 👍 
Didn't notice any change of behaviour on Linux. And on Windows I seem to get one step further now with this PR.
Without PR:
`AttributeError: module 'os' has no attribute 'unsetenv'`
With PR:
`ModuleNotFoundError: No module named 'win32con' etc.` (whatever this means)

PR #205, opened 2020-04-01
api order by ID

will-moore:
```getObjects()``` orders by ID by default

Since we support pagination, we should ensure consistent ordering of objects.
If user doesn't choose any ordering, then we order by ID by default.


joshmoore:
Reopening with #216 merged.

PR #208, opened 2020-04-21
Ignore owner when checking omero.data.dir is accessible

manics:
# What this PR does

The owner/permissions check on the OMERO data dir is too strict, for example it will fail if owned by root even if it is group accessible.

# Testing this PR

1. Create a new OMERO data dir
2. Make it owned by `root`
3. Change the group of the data dir to one that the user that will be running OMERO.server is a member of
4. Make the OMERO data dir group readable/writeable
5. Start the server and it should work!

# Related reading
- https://trello.com/c/qndSyho7/532-omero-admin-datadir-owner-check-is-too-strict

-----
Ported from https://github.com/ome/openmicroscopy/pull/5884
- See also https://forum.image.sc/t/import-data-to-omero-permission-denied/29427/6

### DIFF
--- a/src/omero/plugins/admin.py
+++ b/src/omero/plugins/admin.py
@@ -1593,18 +1593,9 @@ present, the user will enter a console""")
             self.ctx.die(8, "FATAL: OMERO directory does not exist: %s"
                          % pathobj)
 
-        owner = os.stat(filepath)[stat.ST_UID]
-        if owner == 0:
-            msg = ""
-            msg += "FATAL: OMERO directory which needs to be writeable"\
-                " belongs to root: %s\n" % filepath
-            msg += "Please use \"chown -R NEWUSER %s\" and run as then"\
-                " run %s as NEWUSER" % (filepath, sys.argv[0])
-            self.ctx.die(9, msg)
-        else:
-            if not os.access(filepath, mask):
-                self.ctx.die(10, "FATAL: Cannot access %s, a required"
-                             " file/directory for OMERO" % filepath)
+        if not os.access(filepath, mask):
+            self.ctx.die(10, "FATAL: Cannot access %s, a required"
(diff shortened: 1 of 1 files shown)


mtbc:
```
$ omero config get omero.data.dir
/home/mtbc/var/omero/
$ ls -ld /home/mtbc/var/omero/
drwxrwxr-x 2 root mtbc 4096 Apr 22 11:03 /home/mtbc/var/omero/
$ ls /home/mtbc/var/omero/
$ omero admin start
FATAL: OMERO directory which needs to be writeable belongs to root: /home/mtbc/var/omero/
Please use "chown -R NEWUSER /home/mtbc/var/omero/" and run as then run /tmp/omero/bin/omero as NEWUSER
$ pip install -U /tmp/omero-py-5.6.3.dev0.tar.gz 
...
Successfully installed omero-py-5.6.2
$ omero admin start
Creating /tmp/OMERO.server-5.6.0-162-891359e-ice36-b418/var/master
Initializing /tmp/OMERO.server-5.6.0-162-891359e-ice36-b418/var/log
Creating /tmp/OMERO.server-5.6.0-162-891359e-ice36-b418/var/registry
No descriptor given. Using etc/grid/default.xml
Waiting on startup. Use CTRL-C to exit
$ ls -l /home/mtbc/var/omero/
total 16
drwxr-xr-x 2 mtbc mtbc 4096 Apr 22 11:27 BioFormatsCache
drwxr-xr-x 2 mtbc mtbc 4096 Apr 22 11:28 Files
drwxr-xr-x 2 mtbc mtbc 4096 Apr 22 11:27 FullText
drwxr-xr-x 3 mtbc mtbc 4096 Apr 22 11:28 ManagedRepository
$ 
```
Import, web, etc. all work fine.

PR #216, opened 2020-06-02
Attempt to fix Travis builds

manics:
Upgrades all components

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ jobs:
         on:
           tags: true
 
-install: pip3 install --upgrade --upgrade-strategy=eager pip wheel pytest tox
+install: pip3 install --upgrade pip wheel pytest tox virtualenv
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


joshmoore:
@manics : can you explain your thinking on the eager upgrade strategy? Is this something where we should be testing both variants?

manics:
It was suggested as a workaround for Travis failures in https://github.com/jaraco/skeleton/commit/172c52b03a77e70e92a1ee491b059464e938fde3

joshmoore:
Did you try limiting this to just `virtualenv`? I don't know the graph of what these 4 libraries will pull in, and I certainly don't use `upgrade-strategy` regularly enough (tip: I didn't know it existed) to ensure that what we're testing in travis would match what a user would see.

manics:
Just tried it in the last commit, seems to work

joshmoore:
LGTM :+1:

PR #218, opened 2020-06-04
Warn if omero.db.poolsize is not set.

mtbc:
Fixes https://github.com/ome/omero-server/issues/84 by printing the suggested message during `omero admin start` if `omero.db.poolsize` has not been configured.

### DIFF
--- a/src/omero/plugins/admin.py
+++ b/src/omero/plugins/admin.py
@@ -773,6 +773,15 @@ present, the user will enter a console""")
             self.checkwindows(args)
         self.check_lock(config)
 
+        try:
+            config['omero.db.poolsize']
+        except KeyError:
+            self.ctx.out(
+                "WARNING: Your server has not been configured for production "
+                "use. See https://docs.openmicroscopy.org/omero/latest/"
+                "sysadmins/server-performance.html?highlight=poolsize "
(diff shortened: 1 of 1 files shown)


joshmoore:
In general, :100:

However, for me this prints out as:

```
(omero) /opt/omero-py $omero admin start
WARNING: Your server has not been configured for production use. See https://docs.openmicroscopy.org/omero/latest/sysadmins/server-performance.html?highlight=poolsize for more information.
No descriptor given. Using etc/grid/default.xml
Waiting on startup. Use CTRL-C to exit
```

I wonder if there won't be other similar items to check for (certificates?) in which case we might not want an entire block printed at the top.

joshmoore:
NB: I could see getting a quick release with this included.

mtbc:
Would it be better differently formatted, or maybe printed later?

joshmoore:
For the moment, I think having at the top is fine. I'd just make it blockier.

mtbc:
Now updated to:
```
WARNING: Your server has not been configured for production use.
See https://docs.openmicroscopy.org/omero/latest/sysadmins/server-performance.html?highlight=poolsize
for more information.
```

PR #219, opened 2020-06-15
setup.py: do not install Windows batch scripts on non-Windows systems.

carandraug:
The install script currently installs all files under bin. However, with the exception of omero itself, they're all batch scripts for windows only. This PR will only install them under Windows.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -170,6 +170,11 @@ sys.path.append("target")
 from omero_version import omero_version as ov  # noqa
 
 
+scripts = [os.path.join("bin", "omero")]
+if os.name in ("nt", "ce"):
+    scripts += glob.glob(os.path.join("bin", "*.bat"))
+
+
 def read(fname):
     """
     Utility function to read the README file.
@@ -205,7 +210,7 @@ setup(
         'omero.gateway': ['pilfonts/*'],
         'omero.gateway.scripts': ['imgs/*']},
     py_modules=packageless,
-    scripts=glob.glob(os.path.sep.join(["bin", "*"])),
+    scripts=scripts,
     python_requires='>=3',
     install_requires=[
         'future',


joshmoore:
Thanks for keeping us clean, @carandraug.

manics:
Unfortunately `setup.py` is only executed during a build, not at install time, so any published packages will match the state of the machine used to build it. You'll probably need to replace the scripts with an entrypoint console_script instead: https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point

carandraug:
>[...], so any published packages will match the state of the machine used to build it.

If the build should be different on different OS, should there not be different builds for each? But since omero-py is python code only, what's the advantage of having a wheel for it?

> You'll probably need to replace the scripts with an entrypoint console_script instead: [...]

I can merge the `bin/omero` into `src/omero/cli.py` (seems like the best place), but I have no idea when the windows batch scripts are meant to be used or even be able to test them.

manics:
Wheels are quicker to install than source packages:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels
In addition this also means we've standardised our Python packaging and CI across all repositories- we build wheels for all repositories.

We could have platform dependent wheels, but if we're making changes my view is that we should use this as an opportunity to reduce the technical debt and switch to an entrypoint for the `omero` executable (pip will automatically create a platform dependent wrapper when it's installed, negating the need for `omero.bat`).

@joshmoore are `setpythonpath.bat` and `winconfig.bat` old files that can be deleted?

joshmoore:
> @joshmoore are `setpythonpath.bat` and `winconfig.bat` old files that can be deleted?

Yes. Double checked and they were only for Windows server deployments.

manics:
@carandraug I had a go in https://github.com/ome/omero-py/pull/229

PR #221, opened 2020-06-16
Update README release commands to sign release tags

sbesson:
Minor improvements from my experience releasing omero-py 5.7.0.

A second minor is that if the previous dev version does not match the expected release version (in my case `5.6.3.dev0` was released as `5.7.0`), the default command will obviously create an undesired tag/commit. Options are either:
- to review the current version and have a preliminary bump of `bumpversion --no-tag [major|minor|patch]`
- to override the content of `.bumpversion.cfg` using  `bumpversion release --current-version/--new-version
No strong preference either way but if people have a inclination towards either of these options, I can certainly add a sentence to the README as part of this PR

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -2,6 +2,7 @@
 current_version = 5.7.1.dev0
 commit = True
 tag = True
+sign_tags = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\.(?P<release>[a-z]+)(?P<build>\d+))?
 serialize = 
 	{major}.{minor}.{patch}.{release}{build}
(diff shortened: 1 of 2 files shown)


joshmoore:
Doh! Somewhere along the line we lost this:

```
(base) /opt $grep sign */.bumpversion.cfg
ome-zarr-py/.bumpversion.cfg:sign_tags = True
omero-dropbox/.bumpversion.cfg:sign_tags = True
omero-figure/.bumpversion.cfg:sign_tags = True
omero-signup/.bumpversion.cfg:sign_tags = True
```

sbesson:
Updated to set the defaults in the configuration file instead.

joshmoore:
Thanks. I've blindly been assuming I've had this active _everywhere_.

PR #224, opened 2020-06-24
UTF-8 Fixes

joshmoore:
* [x] `omero hql "from Mask"` fails
 * [ ] ~`omero upload` fails (see image.sc)~

### DIFF
--- a/src/omero/cli.py
+++ b/src/omero/cli.py
@@ -530,7 +530,10 @@ class Context(object):
                 raise
         except:
             print("Error printing text", file=sys.stderr)
-            print(text, file=sys.stdout)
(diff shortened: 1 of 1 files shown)


joshmoore:
I've not been able to reproduce the error from https://forum.image.sc/t/omero-deconvolution-recommendation/39047/6?u=joshmoore with:

 - `omero upload`
 - `omero script upload`
 - `omero script launch`

of:

 - files _containing_ unicode
 - files with unicode in the _name_

I'm going to go ahead and get this PR in to match https://github.com/ome/omero-py/pull/236 while we try to reproduce.

PR #226, opened 2020-06-26
fix logging attribute error

### LINKED ISSUE #225: Potential AttributeError in SessionsStore.create
I'm not sure what the best way to fix this is (without potentially breaking other stuff you depend on), but there is an assumption being made in `SessionsStore.create()` that a `self.ctx` attribute has been injected into the object "somewhere" else:

https://github.com/ome/omero-py/blob/8d711f577d1000aad49ea794324b49450c952878/src/omero/util/sessions.py#L406-L410

If you create an unowned store, and try `store.create()` but hit an exception when [submitting the `TimeoutRequest`](https://github.com/ome/omero-py/blob/8d711f577d1000aad49ea794324b49450c952878/src/omero/util/sessions.py#L404), then the exception is not handled correctly.

tlambert03:
one possible solution that fixes #225 but still lets you log from `self.ctx` if you want.  Or just let me know if you want to ditch self.ctx altogether.

### DIFF
--- a/src/omero/util/sessions.py
+++ b/src/omero/util/sessions.py
@@ -406,9 +406,7 @@ class SessionsStore(object):
             except Exception:
                 import warnings
 
-                warnings.warn(
-                    f"Invalid session timeout ({timeout}) requested when creating session."
-                )
+                warnings.warn("Invalid session timeout requested when creating session.")
 
             # Reload session
             sess = sf.getSessionService().getSession(uuid)
--- a/src/omero/util/sessions.py
+++ b/src/omero/util/sessions.py
@@ -404,12 +404,11 @@ class SessionsStore(object):
                 cb = client.submit(req)  # Response is "OK"
                 cb.close(True)
             except omero.CmdError as ce:
-                _logger = self.ctx.dbg if hasattr(self, 'ctx') else self.logger.debug
-                _logger(str(ce.err))
+                self.logger.warn(str(ce.err))
             except:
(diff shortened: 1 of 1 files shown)


joshmoore:
I'd err on the side of ditching. This class should really have no concept of `self.ctx` (That's a purely CLI thing.) Looking at this, I think this should also be a warn. The user has requested a timeout, and if that can't be set (or is incorrect) it seems like at least a WARN. (Perhaps the exception should even be thrown, but I'll need to review more for that.)

tlambert03:
k, happy to convert to a warn or raise.  just let me know.

joshmoore:
Let's go with a warn and get it released, then I can follow up with the rest.

tlambert03:
k, do you want the formatted traceback in the warning?

joshmoore:
Sorry for being a pain, @tlambert03, but can we go back to 2d34519dd2d7338b0bf5e47762ff6c4988653175 and:

 - keep the CmdError / Exception split,
 - use logger.warn for the CmdError at `WARN`, and
 - use logger.error for the catch-all Exception?

tlambert03:
ok have another look at your convenience.  if you also wanted the `self.ctx` check added back, lemme know.

joshmoore:
> ok have another look at your convenience.

Awesome, thank you!


> if you also wanted the `self.ctx` check added back, lemme know.

Nah. That should have never happened. :+1:

PR #228, opened 2020-06-30
Add PyYAML as a dependency of omero-py

sbesson:
In the current construction of omero-py, there are two public APIs which require PyYAML at runtime: the CLI `import --bulk` functionalitity as well as the `omero.util.pydict_text_io` utility consumed by a few CLI plugins (render, metadata) for loading YAML files.

497f1bd makes `PyYAML` a mandatory dependency of `omero-py`
dcde6b2 also removes the optional `import yaml` handling and replaces usage of the deprecated `yaml.load` and `yaml.load_all` by `yaml.safe_load` and `yaml.safe_load_all`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -211,6 +211,7 @@ setup(
         'future',
         'numpy',
         'Pillow',
+        'PyYAML',
         'zeroc-ice>=3.6.4,<3.7',
         'pywin32; platform_system=="Windows"',
     ],
--- a/src/omero/util/pydict_text_io.py
+++ b/src/omero/util/pydict_text_io.py
@@ -28,21 +28,14 @@ import json
 import re
 from omero.rtypes import unwrap
 from future.utils import bytes_to_native_str
-
-try:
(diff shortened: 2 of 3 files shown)


manics:
In case you missed it https://github.com/sbesson/omero-py/pull/1

joshmoore:
The changes all look good to me. Testing today was all green (and travis is happy). Merging.

sbesson:
I realize I haven't discussed versioning. My inclination would be to release by incrementing the patch version i.e. `5.7.2` as the API is not modified and the runtime dependency exception has always felt as a bug to me.

joshmoore:
I was thinking about that. I don't imagine adding a dependency, now that it happens transitively, requires a minor bump. Though now that I think about it, I wonder if anything else like https://github.com/ome/conda-omero-py/blob/master/meta.yaml needs a bump, too. We probably need to codify the steps on a requirement addition/modification.

PR #230, opened 2020-07-01
hql.py: when querying masks, filter `bytes` field

joshmoore:
`omero hql "select s from Shape s"` or similar fails with

```
UnicodeEncodeError: 'utf-8' codec can't encode character '\udcb0' in position 3797: surrogates not allowed
```

due to the bytes field.


joshmoore:
This is the only field with that name:

```
(cd /opt/omero-build/omero-model/src/main/resources/mappings/; git grep bytes)
roi.ome.xml:            <optional name="bytes" type="byte[]"/>
```

PR #231, opened 2020-07-01
Activate test_import unit tests with OMERO_USERDIR

manics:
This is a followup to https://github.com/ome/omero-py/pull/162 (the last ~3~ 4 commits in this PR are new).
1. If `OMERO_USERDIR` is not set this uses a temporary directory to ensure the unit test does not modify the user's environment since jars are automatically downloaded.
2. Activates all import tests that were previously disabled when `OMERODIR` wasn't set
3. Fixes broken import tests which apparently were never run in CI
4. Disables the tests activated in step 2 on Windows only since they're failing

Downside of course is that anyone running the full test suite without setting `OMERO_USERDIR` will end up downloading OMERO.java once (the fixture that sets `OMERO_USERDIR` to a tempdir is scoped to the session not the function), but I think this is preferable to modifying the default `OMERO_USERDIR`. Alternatively we could disable the tests by default, but then they'll probably break again.

### DIFF
--- a/src/omero/plugins/import.py
+++ b/src/omero/plugins/import.py
@@ -32,11 +32,23 @@ from past.builtins import basestring
 from builtins import object
 import os
 import csv
+from shutil import copyfileobj
 import sys
 import shlex
 
+try:
+    from io import BytesIO
+    from urllib.request import urlopen
+except ImportError:
+    # Python 2
+    from StringIO import StringIO as BytesIO
+    from urllib import urlopen
+from zipfile import ZipFile
+
+
 from omero.cli import BaseControl, CLI
 import omero.java
+from omero.util import get_omero_userdir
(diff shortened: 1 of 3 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#364](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/364/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/364/consoleText) for more details.
Possible conflicts:
  - PR #228 sbesson 'Add PyYAML as a dependency of omero-py'
    - test/unit/clitest/test_import.py

--conflicts

manics:
These commits are included in https://github.com/ome/omero-py/pull/162

Issue #210, opened 2020-05-06
OMERO CLI fails if a file ~/omero exists


manics:
```
[test@1f99c18b86c3 ~]$ pwd
/home/test

[test@1f99c18b86c3 ~]$ ls

[test@1f99c18b86c3 ~]$ ln -s /opt/omero/server/venv3/bin/omero

[test@1f99c18b86c3 ~]$ ./omero login
WARNING:omero.util.TempFileManager:Invalid tmp dir: /home/test/omero/tmp
Traceback (most recent call last):
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/util/temp_files.py", line 178, in tmpdir
    self.create(target)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero/util/temp_files.py", line 246, in create
    dir.makedirs(0o700)
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero_ext/path.py", line 1254, in makedirs
    os.makedirs(self, mode)
  File "/usr/lib64/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
NotADirectoryError: [Errno 20] Not a directory: path('/home/test/omero/tmp')
Could not access session dir: /home/test/omero/sessions
```

joshmoore:
The use of `~/omero` has bothered me for some time. I wonder if it's worth using this as a chance to move something more platform-appropriate: `~/.config/omero` + `$TMPDIR/$USER-omero` + ..., etc.

manics:
What do you think of using https://pypi.org/project/appdirs/ ?

joshmoore:
> - Project officially classified as Stable which is important for inclusion in other distros such as ActivePython.
> - First of several incremental releases to catch up on maintenance.

;)

PR #233, opened 2020-07-09
Replace ~/omero with standard OS-dependent paths

manics:
Replaces `~/omero` with "standard" paths according to each operating system: https://pypi.org/project/appdirs/

Closes https://github.com/ome/omero-py/issues/210

### DIFF
--- a/src/omero/util/__init__.py
+++ b/src/omero/util/__init__.py
@@ -12,6 +12,7 @@ from builtins import str
 from future.utils import native_str
 from past.utils import old_div
 from builtins import object
+from appdirs import user_data_dir, user_cache_dir
 import os
 import sys
 import Ice
@@ -29,6 +30,8 @@ import uuid
 import omero.ObjectFactoryRegistrar as ofr
 
 from omero.util.decorators import locked
+from omero_version import omero_version
+
 import omero_ext.path as path
 
 LOGDIR = os.path.join("var", "log")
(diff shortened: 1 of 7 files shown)


src/omero/util/__init__.py
```diff
@@ -42,6 +45,9 @@
 orig_stdout = sys.stdout
 orig_stderr = sys.stderr
 
+# Application name, Owner (Windows only), Major version
+APPDIR_DEFAULTS = ('OMERO.cli', 'OME', omero_version.split('.')[0])
```
manics, line 49: On OS X this results in:
- `~/Library/Application Support/OMERO.cli/5
- `~/Library/Caches/OMERO.cli/5

manics:
Is this considered a breaking change that should wait for 6.0.0?. If so this could be split so only `omero.util.get_omero_user_cache_dir()` which was added in #162 uses the new paths

sbesson:
Few immediate thoughts:

- agreed the changes are not breaking the API although we might want to review the impact of such a change
-  I suspect we will need to consider the case where downstream applications are accessing application data and moving away from `~/omero` would break the expectations. In that sense, I think the delineation proposed in https://github.com/ome/omero-py/pull/233#issuecomment-656303819 is probably an acceptable compromise as anything stored in a cache directory should be considered as potentially transient
- more generally, do you have some mechanism that would allow the migration of data between the old user/cache directory and the new one
- the application specific aspect of the paths also raises the questions of the unification with other OMERO clients e.g. OMERO.insight is using `~/omero/log`

manics:
I'll rebase on top of https://github.com/ome/omero-py/pull/240

PR #239, opened 2020-07-30
Always marshal tile metadata on presence of pyramid

chris-allan:
It is possible, in particular for floating point data, that a pyramid is
present even if the server thinks a pyramid is not required.  The
condition for tile metadata being available is now pyramid presence.

See:

 * ome/omero-romio#24
 * ome/omero-web#193


chris-allan:
Additional requirement for restoring all the pyramid metadata that was started in ome/omero-web#193.

/cc @sbesson

joshmoore:
The change seems straight-forward enough, though imagine we should clarify in https://github.com/ome/omero-blitz/blob/master/src/main/slice/omero/api/PyramidService.ice when each of the methods should be used.

chris-allan:
@joshmoore: Yeah, in particular `requirePixelsPyramid()` as currently documented is already largely incorrect:

 * https://github.com/ome/omero-blitz/blob/master/src/main/slice/omero/api/PyramidService.ice#L38

PR #242, opened 2020-08-11
Use OS specific application directories instead of ~/omero

manics:
This is https://github.com/ome/omero-py/pull/233 rebased and modified to only use the `user_cache_dir` property. `get_omero_userdir()` is unchanged (still returns `~/omero`) but this can be changed to `user_data_dir` in 6.0.0 when breaking changes are allowed.

You can enable the new behaviour for `get_omero_userdir()` by setting `OMERO_USERDIR=""` (empty string, as opposed to not setting it).

Related to https://github.com/ome/omero-py/issues/210 but does not fully resolve it- need to wait until the new behaviour is the default

No migration options are supported since the cache dir has not yet been released. Longer term I'm inclined not to, AFAIK at the moment sessions are the only meaningful state stored under `~/omero`, and since a major release probably implies a major release of the server and therefore a restart there's no point migrating an old session.

Decisions:
- [x] Include the major version (currently `5`, will be `6` in future) in the directory path or not.

### DIFF
--- a/src/omero/util/__init__.py
+++ b/src/omero/util/__init__.py
@@ -839,12 +839,20 @@ def get_user(default=None):
 
 
 def get_omero_userdir():
-    """Returns the OMERO user directory"""
+    """
+    Returns the OMERO user directory
+
+    In 6.0.0 the default will change to use appdirs.user_data_dir.
+    You can enable this behaviour now by setting OMERO_USERDIR="" (empty
+    string) instead of unset.
+    """
     omero_userdir = os.environ.get('OMERO_USERDIR', None)
     if omero_userdir:
         return path.path(omero_userdir)
-    else:
+    elif omero_userdir == "":
         return path.path(user_data_dir(*APPDIR_DEFAULTS))
+    else:
+        return old_div(path.path(get_user_dir()), "omero")
 
 
 def get_omero_user_cache_dir():
@@ -861,7 +869,20 @@ def get_omero_user_cache_dir():
 
 
 def get_user_dir():
-    return get_omero_userdir()
+    exceptions_to_handle = (ImportError)
+    try:
+        from pywintypes import com_error
+        from win32com.shell import shellcon, shell
+        exceptions_to_handle = (ImportError, com_error)
+        homeprop = shell.SHGetFolderPath(0, shellcon.CSIDL_APPDATA, 0, 0)
+    except exceptions_to_handle:
+        homeprop = os.path.expanduser("~")
+
+    if "~" == homeprop:
+        # ticket:5583
+        raise Exception("Unexpanded '~' from expanduser: see ticket:5583")
+
+    return homeprop
 
 
 def edit_path(path_or_obj, start_text):
--- a/src/omero/util/__init__.py
+++ b/src/omero/util/__init__.py
@@ -46,7 +46,7 @@ orig_stdout = sys.stdout
 orig_stderr = sys.stderr
 
 # Application name, Author (Windows only), Major version
-APPDIR_DEFAULTS = ('OMERO.cli', 'OME', omero_version.split('.')[0])
+APPDIR_DEFAULTS = ('OMERO.py', 'OME', omero_version.split('.')[0])
 
 
 def make_logname(self):
--- a/src/omero/util/__init__.py
+++ b/src/omero/util/__init__.py
@@ -845,6 +845,8 @@ def get_omero_userdir():
     In 6.0.0 the default will change to use appdirs.user_data_dir.
     You can enable this behaviour now by setting OMERO_USERDIR="" (empty
     string) instead of unset.
+
+    Note that files in the old user directory will not be migrated.
     """
     omero_userdir = os.environ.get('OMERO_USERDIR', None)
     if omero_userdir:
--- a/src/omero/util/__init__.py
+++ b/src/omero/util/__init__.py
@@ -45,7 +45,7 @@ LOGMODE = "a"
 orig_stdout = sys.stdout
 orig_stderr = sys.stderr
 
-# Application name, Owner (Windows only), Major version
+# Application name, Author (Windows only), Major version
 APPDIR_DEFAULTS = ('OMERO.cli', 'OME', omero_version.split('.')[0])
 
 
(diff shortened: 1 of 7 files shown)


test/unit/test_util.py
```diff
@@ -314,3 +314,20 @@ def test_paste_image(self):
         data_canvas[256, 256] = [255, 255, 0]
         canvas = Image.fromarray(data_canvas, 'RGB')
         image_utils.paste_image(img, canvas, 0, 0)
+
+
+class TestUserdirs(object):
+
+    def testUserdirEnvironmentDefault(self, monkeypatch):
```
jburel, line 321: The returned value of ``get_omero_user_cache_dir`` has changed so it will be good to add test for that too

manics: Added

jburel:
if the user wants to test the new feature, anything stores in the current selected location will be not be moved since there is no migration is offered, maybe the doc should indicate that.
Insight is in maintenance mode, but we should probably unify that cc @dominikl

manics:
Done.

A bit of googling found https://github.com/harawata/appdirs

jburel:
insight issue created https://github.com/ome/omero-insight/issues/160

Thanks for the change.

src/omero/util/__init__.py
```diff
@@ -42,6 +45,9 @@
 orig_stdout = sys.stdout
 orig_stderr = sys.stderr
 
+# Application name, Owner (Windows only), Major version
+APPDIR_DEFAULTS = ('OMERO.cli', 'OME', omero_version.split('.')[0])
```
joshmoore, line 49: I'm not sure about the use of `cli` here. I could see an argument for `py` as well, or even for no suffix to have both Python and Java sharing sessions.

joshmoore: Guess I'm not sure what "Owner" means either.

joshmoore (commented):
One question for everyone re: `OMERO.cli`. Otherwise, seems straight-forward enough:

```
bash-4.2$ .venv3/bin/python -m "omero.util.temp_files" dir
/home/omero/omero/tmp/omero_omero/2432
bash-4.2$ OMERO_USERDIR="" .venv3/bin/python -m "omero.util.temp_files" dir
/home/omero/.local/share/OMERO.cli/5/tmp/omero_omero/2434
```

src/omero/util/__init__.py
manics: Not sure why I put owner, it should be `author` e.g. `OME`, `Mozilla`, `Microsoft`, etc. Only relevant on Windows https://github.com/ActiveState/appdirs/blob/1.4.4/appdirs.py#L50-L53

manics: I can see arguments either way for including/excluding the major version, using plain OMERO vs OMERO.component. I guess we'll just have to pick one. OTOH if we include the major version we've got the option of changing from OMERO.component to OMERO in version 6.

sbesson: Small preference for `OMERO.py` unless we have a good feeling about the ability for sharing sessions with Java in which case I can see testing `OMERO`.

Alternatively, is it possible to make this configurable so that we can test the sharing later?

sbesson: Fair enough

```suggestion
APPDIR_DEFAULTS = ('OMERO.py', 'OME', omero_version.split('.')[0])
```

test/unit/test_util.py
```diff
@@ -314,3 +315,33 @@ def test_paste_image(self):
...
+        monkeypatch.delenv('OMERO_USERDIR', raising=False)
+
+        assert get_omero_userdir().basename() == 'omero'
+
+        c = get_omero_user_cache_dir()
+        assert c.basename() == omero_version.split('.')[0]
+        assert 'OMERO.cli' in str(c)
```
manics, line 329:
```suggestion
        assert 'OMERO.py' in str(c)
```

test/unit/test_util.py
```diff
@@ -314,3 +315,33 @@ def test_paste_image(self):
...
+
+    def testUserdirEnvironmentAppdir(self, monkeypatch):
+        monkeypatch.setenv('OMERO_USERDIR', '')
+
+        d = get_omero_userdir()
+        assert d.basename() == omero_version.split('.')[0]
+        assert 'OMERO.cli' in str(d)
```
manics, line 336:
```suggestion
        assert 'OMERO.py' in str(d)
```

joshmoore (approved):
:+1: can certainly live with the use of "OMERO.py" for the moment which will at least cover the napari use case. I don't know of anyone who has made use of those files from Java.

ehrenfeu:
Hi all, especially @manics and @sbesson 

I'd like to point out that (if I understood correctly) with the changes planned for 6.0.0 in [`get_omero_userdir()`][1] it will be difficult to e.g. run an `omero import` if the system user doesn't have write permissions to its `$HOME` directory.

At least I haven't found a way to influence [appdirs][2] choice on the path, e.g. through an environment variable.

This might sound esoteric, but it's affecting me in [hrm-omero][3] as the way the code usually gets called on an HRM installation is through the system user running the web server (e.g. `www-data`) and quite often (e.g. on CentOS 7) that one is not allowed to write to `/var/www/` or similar. In that case the download and extraction of `OMERO.java.zip` fails (silently) and the import process breaks.

Would it be possible to have an optional way to influence where that stuff ends up in the file system?

Cheers,
Niko

[1]: https://github.com/manics-archive/omero-py/blob/09dddcfd3c55347b383a425a1fa3c5e2a2366b67/src/omero/util/__init__.py#L839-L857
[2]: https://pypi.org/project/appdirs/
[3]: https://github.com/imcf/hrm-omero

will-moore:
I don't see any other changes planned for 6.0.0 that corresponds to what's in the docstring `In 6.0.0 the default will change to use appdirs.user_data_dir`, so maybe we should open a PR to do that?

But even in that case, you can still override by setting `OMERO_USERDIR`.

ehrenfeu:
Thanks @will-moore - if setting `OMERO_USERDIR` will continue to work I am all fine.

It's just that I read the "*In 6.0.0 the default will change to use appdirs.user_data_dir*" that the `appdirs.user_data_dir` will have priority over the `OMERO_USERDIR` variable. That's what got me worried.

PR #246, opened 2020-08-24
admin.py: optionally check for system manager environment variable

manics:
Adds a config property that can be set to the name of an environment variable that is only set when running under systemd or some other service manager. `omero admin start|stop` will check this variable and abort if it's not set.

## Example using systemd
If you have systemd v232+ the `INVOCATION_ID` environment variable should be set by systemd so you can check for this https://serverfault.com/a/927481
```
$ cat /opt/omero/server/config/10-systemd.omero
config set omero.server.servicemanager.checkenv INVOCATION_ID
```

If you have an older version of systemd you can define your own non-empty variable in your systemd service file:
```diff
[Service]
User=omero-server
UMask=0002
Type=forking
PIDFile=/opt/omero/server/OMERO.server/var/master/master.pid
Restart=no
RestartSec=10
+Environment=CHECK_FOR_THIS_VARIABLE=something
# Allow up to 5 min for start/stop
TimeoutSec=300
ExecStartPre=/opt/omero/server/OMERO.server/bin/omero load --glob /opt/omero/server/config/*.omero
ExecStart=/opt/omero/server/OMERO.server/bin/omero admin start
ExecStop=/opt/omero/server/OMERO.server/bin/omero admin stop
```

```
$ cat /opt/omero/server/config/10-systemd.omero
config set omero.server.servicemanager.checkenv CHECK_FOR_THIS_VARIABLE
```

If you then attempt to start/stop omero outside systemd:
```
[vagrant@omero ~]$ sudo systemctl start omero-server
[vagrant@omero ~]$ sudo -u omero-server /opt/omero/server/OMERO.server/bin/omero admin stop
ERROR: OMERO is configured to run under a service manager but CHECK_FOR_THIS_VARIABLE is not set
[vagrant@omero ~]$ sudo -u omero-server /opt/omero/server/OMERO.server/bin/omero admin restart
ERROR: OMERO is configured to run under a service manager but CHECK_FOR_THIS_VARIABLE is not set
```

Note since the omero configuration is updated from `/opt/omero/server/config/*.omero` you must've started omero at least once with systemd after updating the configuration.

I'll document this property in https://github.com/ome/openmicroscopy/blob/develop/etc/omero.properties if everyone is happy with this.

Includes a test, though I had to put it in a new file because `test_admin.py` contains a global autouser fixture that requires `OMERODIR`, which means it's not tested on Travis.

A similar property can be added to `omero web`. Since these are decoupled I don't think they should share the same property, instead it would be `omero.web.servicemanager.checkenv`.

@stick 

Note this was updated so the property is now `omero.server.servicemanager.checkenv`, not `omero.admin.servicemanager.checkenv`

### DIFF
--- a/src/omero/plugins/admin.py
+++ b/src/omero/plugins/admin.py
@@ -721,6 +721,20 @@ present, the user will enter a console""")
                          % os.path.sep.join(["etc", "grid", __d__]))
         return descript
 
+    def _requires_service_manager(self, config):
+        """
+        Checks whether OMERO is being managed by a service manager by
+        checking that a specified environment variable is non-empty
+        """
+        service_env = config.as_map().get(
+            "omero.admin.servicemanager.checkenv")
+        if service_env:
+            service_envvalue = os.getenv(service_env)
+            if not service_envvalue:
+                self.ctx.die(112, (
+                    "ERROR: OMERO is configured to run under a service "
+                    "manager but {} is not set".format(service_env)))
+
     def checkwindows(self, args):
         r"""
         Checks that the templates file as defined in etc\Windows.cfg
@@ -760,6 +774,7 @@ present, the user will enter a console""")
         First checks for a valid installation, then checks the grid,
         then registers the action: "node HOST start"
         """
+        self._requires_service_manager(config)
         self.check_access(mask=os.R_OK, config=config)
         self.checkice()
         self.check_node(args)
@@ -974,6 +989,7 @@ present, the user will enter a console""")
         """
         Returns true if the server was already stopped
         """
+        self._requires_service_manager(config)
         self.check_node(args)
         if args.force_rewrite:
             self.rewrite(args, config, force=True)
--- a/src/omero/plugins/admin.py
+++ b/src/omero/plugins/admin.py
@@ -42,6 +42,7 @@ from omero.cli import (
     DiagnosticsControl,
     DirectoryType,
     NonZeroReturnCode,
+    ServiceManagerMixin,
     UserGroupControl,
 )
 
@@ -93,7 +94,11 @@ if platform.system() == 'Windows':
 
 class AdminControl(DiagnosticsControl,
                    WriteableConfigControl,
-                   UserGroupControl):
+                   UserGroupControl,
+                   ServiceManagerMixin):
+
+    # Require by omero.cli.ServiceManagerMixin
+    SERVICE_MANAGER_KEY = 'server'
 
     def _complete(self, text, line, begidx, endidx):
         """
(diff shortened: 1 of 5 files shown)


src/omero/plugins/admin.py
```diff
@@ -721,6 +721,20 @@ def _descript(self, args):
                          % os.path.sep.join(["etc", "grid", __d__]))
         return descript
 
+    def _requires_service_manager(self, config):
```
joshmoore, line 724: If this should also apply to omero-web, I could see this in a mixing superclass like the other superclasses for the AdminControl.

manics: Are you thinking something in https://github.com/ome/omero-py/blob/master/src/omero/cli.py since that's where `DiagnosticsControl` is, e.g. `ServiceManagerControl`?

manics: Note the property name would be different so it would have to be a parameter.

joshmoore:
```
class AdminControl(ServiceManagerControl, ...):
    SERVICE_MANAGER_KEY = "server"
```
or whatever, yeah.

joshmoore:
Barring the introduction of the new class name suffix (`Mixin`) which I may have caused, this generally looks good to me. I haven't had a change to test yet. What would be your first targets for this?

manics:
I went for a mixin because it only provides a utility method for internal plugin use, it didn't seem right to construct a full Control object. Need to decide between consistency vs gradual improvement.

Once this is released I'd start by adding this to our OMERO playbooks through [these role vars](https://github.com/ome/ansible-role-omero-server):
- Add a required variable in `omero_server_systemd_environment` e.g. `OMERO_SYSTEMD_MANAGED=1`
- Add the corresponding property to `omero_server_config_set` e.g. `config set omero.admin.servicemanager.checkenv OMERO_SYSTEMD_MANAGED`

And similarly for OMERO.web

If there are no unforeseen issues we can consider making this mandatory in a future release of the roles.

will-moore:
Does ```$ sudo systemctl start omero-server``` set or unset the environment variable? or neither?

Does ```bin/omero admin stop``` fail because you started via `systemctl` or because `CHECK_FOR_THIS_VARIABLE` is not set in that environment?
```
[vagrant@omero ~]$ sudo -u omero-server /opt/omero/server/OMERO.server/bin/omero admin stop
ERROR: OMERO is configured to run under a service manager but CHECK_FOR_THIS_VARIABLE is not set
```

If you did set ```$ export CHECK_FOR_THIS_VARIABLE=foo``` would `$ omero admin stop` then work?
That's kinda how the ERROR reads.

manics:
The latter. There's no easy way to tell whether omero was started by systemd (or any other service manager), so the idea of this workaround is to set a variable in the systemd config that wouldn't otherwise be set in a normal environment. You're correct that if you manually set that variable you'd bypass the check in `omero admin`.

will-moore:
How about:
```
ERROR: OMERO is configured to run under a service manager which should also set CHECK_FOR_THIS_VARIABLE
```

will-moore:
Good, thanks.

manics:
If everyones happy with this it'd be nice to include it in https://github.com/ome/omero-py/pull/248 but it's not essential.

joshmoore:
> If everyones happy with this

WFM. :+1:

PR #247, opened 2020-08-26
Link to source on pypi homepage

manics:
The [current pypi page](https://pypi.org/project/omero-py/) doesn't link to the source.
This changes the project URL to the source, and adds the current URL as [documentation](https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -199,7 +199,9 @@ setup(
     url='https://github.com/ome/omero-py',
     project_urls={
         'Documentation':
-            'https://docs.openmicroscopy.org/latest/omero/developers',
+            'https://docs.openmicroscopy.org/omero/latest/developers/'
+            'Python.html',
+        'Bug tracker': 'https://github.com/ome/omero-py/issues',
     },
     package_dir={"": "target"},
     packages=packages,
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -198,7 +196,11 @@ def read(fname):
         # http://pypi.python.org/pypi?%3Aaction=list_classifiers
     author="The Open Microscopy Team",
     author_email="ome-devel@lists.openmicroscopy.org.uk",
-    url=url,
+    url='https://github.com/ome/omero-py',
+    project_urls={
+        'Documentation':
+            'https://docs.openmicroscopy.org/latest/omero/developers',
```
manics, line 202: Do you think https://docs.openmicroscopy.org/omero/latest/developers/Python.html is more appropriate?

sbesson: :+1: for using the GitHub repo as the home page and adding a documentation link. Looking at https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls, I can certainly see additional fields appearing in the future. Maybe add `[Bug ]Tracker` pointing to the GitHub issues as part of this round?

manics:
With the pending 5.8.0 release we could also fix the email address, currently `devel@lists.openmicroscopy.org.uk`, if anyone has a suggestion?

sbesson (approved):

sbesson:
I have briefly looked into alternatives for the `author_email` metadata following the decommissioning of the mailing lists but have found none I was satisfied with.

Issue #250, opened 2020-09-08
Cope with MinProtocol = TLSv1.2


mtbc:
https://docs.openmicroscopy.org/omero/5.6.2/sysadmins/unix/server-debian10-ice36.html needs to tweak `/etc/ssl/openssl.cnf` and some experimentation suggests that remains true even with this installed. I wonder if some adjustment to this could make it work around out-of-the-box tighter OpenSSL configuration.

jburel:
I will open a PR later on

manics:
One more thing, can you see if `TLS1_0,TLS1_1,TLS1_2` also works? That includes the current protocol, but hopefully the client will just use whichever one(s) it supports

jburel:
I am planning to test that option

jburel:
see https://github.com/ome/omero-py/pull/251

PR #251, opened 2020-09-10
Add support for mutliple Ice TLS protocols

jburel:
The issue was noticed when trying to connect running server on Ubuntu20.04 or CentOS 8
It was possible to connect for example from a mac to Ubuntu20.04 server but not possible to connect to Ubuntu20.04 to Ubuntu20.04 server. See https://travis-ci.org/github/ome/omero-install/jobs/725545346
This PR adds support for multiple protocols

see https://github.com/ome/omero-py/issues/250

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -243,7 +243,8 @@ class BaseClient(object):
 
         self._optSetProp(id, "IceSSL.VerifyDepthMax", "6")
         self._optSetProp(id, "IceSSL.VerifyPeer", "0")
-        self._optSetProp(id, "IceSSL.Protocols", "tls1")
+        self._optSetProp(id, "IceSSL.ProtocolVersionMax", "tls1_2")
+        self._optSetProp(id, "IceSSL.Protocols", "tls1_0,tls1_1,tls1_2")
 
         # Setting block size
         self._optSetProp(
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -243,7 +243,6 @@ class BaseClient(object):
 
         self._optSetProp(id, "IceSSL.VerifyDepthMax", "6")
         self._optSetProp(id, "IceSSL.VerifyPeer", "0")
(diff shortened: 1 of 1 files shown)


mtbc:
:+1: With this (and maybe omero-certificates) we also shouldn't have to fiddle with `/etc/ssl/openssl.cnf` for the Debian 10 installation.

manics:
I think you'll still need omero-certificates on the server side, but hopefully no other server or client config will be required

manics:
I've edited the title so it can go straight into a changelog

jburel:
@mtbc I did not notice problem with debian 10 docker image

mtbc:
https://docs.openmicroscopy.org/omero/5.6.2/sysadmins/unix/server-debian10-ice36.html has a "Fix openssl issues" section.

jburel:
I forgot about that! This is done in the docker image that's true

manics:
So.... we need another docs release?

jburel:
We will need a doc release regardless
cf. https://github.com/ome/omero-install/pull/244

mtbc:
To get the latest commit I did,
```
pip install -U https://codeload.github.com/jburel/omero-py/zip/ice_protocols
```
Together with omero-certificates this lets me log in from CLI but not to import,
```
2020-09-11 09:55:08,093 1144       [      main] INFO       ome.formats.OMEROMetadataStoreClient - Attempting initial SSL connection to 127.0.0.1:4064
2020-09-11 09:55:08,413 1464       [      main] ERROR  formats.importer.cli.CommandLineImporter - Error during import process.
Ice.SecurityException: javax.net.ssl.SSLHandshakeException: Received fatal alert: protocol_version
```
My system is a weird mix of stuff so I wouldn't read too much into that, my point is more that: whatever testing we do for this kind of thing should probably include an import.

jburel:
Using today's omero-py build
https://travis-ci.org/github/ome/omero-install/builds/726206712

I will add an import to testing process via travis

manics:
No objections so merging

PR #252, opened 2020-09-14
Adjust how --include and --exclude are presented as options.

mtbc:
Mirrors commends on https://github.com/ome/omero-cli-duplicate/pull/12 given those were adapted from these.

### DIFF
--- a/src/omero/cli.py
+++ b/src/omero/cli.py
@@ -2026,7 +2026,7 @@ class GraphControl(CmdControl):
             nargs="+", type=lambda s: s.split(","), action="append")
         parser.add_argument(
             "--ordered", action="store_true",
-            help=("Pass multiple objects to commands strictly in the order "
+            help=("Pass objects to commands in the given order"))
                   "given, otherwise group into as few commands as possible."))
         parser.add_argument(
             "--list", action="store_true",
--- a/src/omero/cli.py
+++ b/src/omero/cli.py
@@ -33,7 +33,6 @@ from builtins import input
 from builtins import map
 from builtins import str
 from builtins import range
-from builtins import bytes
 from past.utils import old_div
 from builtins import object
 sys = __import__("sys")
--- a/src/omero/cli.py
+++ b/src/omero/cli.py
@@ -2017,10 +2017,12 @@ class GraphControl(CmdControl):
         self._add_wait(parser, default=-1)
         parser.add_argument(
             "--include",
-            help="Modifies the given option by including a list of classes")
+            help="Modifies the given option by including a list of classes",
+            metavar="CLASS")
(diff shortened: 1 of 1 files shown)


src/omero/cli.py
```diff
@@ -2017,10 +2016,12 @@ def _configure(self, parser):
         self._add_wait(parser, default=-1)
         parser.add_argument(
             "--include",
-            help="Modifies the given option by including a list of classes")
+            help="Specify kinds of object to include",
```
manics, line 2019: The previous help implies a "list of classes", this implies only one. Which is correct?

mtbc:
```python
        if args.include:
            inc = args.include.split(",")
```

mtbc: "kinds" implies multiple, surely?

mtbc:
For testing one can adjust the  `"omero.cmd.graphs"` line in `logback.xml` from INFO to DEBUG to get a full report in `Blitz-0.log` of the options that the CLI sent to the server.

src/omero/cli.py
```diff
@@ -2017,10 +2016,14 @@ def _configure(self, parser):
...
+            nargs="+", type=lambda s: s.split(","), action="append")
         parser.add_argument(
             "--exclude",
-            help="Modifies the given option by excluding a list of classes")
+            help="Specify kinds of object to exclude",
+            metavar="CLASS",
+            nargs="+", type=lambda s: s.split(","), action="append")
         parser.add_argument(
             "--ordered", action="store_true",
             help=("Pass multiple objects to commands strictly in the order "
```
joshmoore, line 2029: This is now the only exceptionally long help line.

mtbc: We could drop the "otherwise"?

mtbc:
```suggestion
            help=("Pass objects to commands in the given order"))
```

joshmoore: Happy to take this suggestion, but it doesn't sufficiently shorten the help:

```
  --exclude CLASS [CLASS ...]    Specify kinds of object to exclude
  --ordered                      Pass multiple objects to commands strictly in the order given, otherwise group into
as few commands as possible.
  --list                         Print a list of all available graph specs
```
so we can also hold off. I imagine what we will need to do long-term is let the class hierarchy define blocks of HELP that get added:

```
f``
Specific help

# Options
{"\n".join(self.other_options)}
``
```
Let's get this in and then we can overall in the future.

cc: @sbesson @manics @will-moore

mtbc: :+1: for getting it in. :smiley:

PR #254, opened 2020-09-17
Fix syntax issue in CLI argument parsing.

mtbc:
See diff.


joshmoore (approved):
That's what I get for not waiting on travis....

PR #256, opened 2020-09-24
Also check stdout is a tty when requiring interactive credentials

sbesson:
Fixes https://github.com/IDR/idr-utils/issues/27

Can be tested with the following minimal workflow

```
omero logout # ensure no session is active
omero login > out # should fail with an error message
omero login # enter valid credentials
omero login > out # should re-use the existing session
```

### DIFF
--- a/test/unit/clitest/test_sess.py
+++ b/test/unit/clitest/test_sess.py
@@ -40,6 +40,8 @@ def istty(monkeypatch):
         return True
     monkeypatch.setattr(sys.stdin, 'isatty', isatty)
     assert sys.stdin.isatty()
+    monkeypatch.setattr(sys.stdout, 'isatty', isatty)
+    assert sys.stdout.isatty()
 
 
 class MyStore(SessionsStore):
--- a/test/unit/clitest/test_sess.py
+++ b/test/unit/clitest/test_sess.py
@@ -410,7 +410,7 @@ class TestSessions(object):
 
     @pytest.mark.parametrize('connection', CONNECTION_TYPES)
     @pytest.mark.parametrize('group', DIFFERENT_GROUPS)
-    def testReuseFromDifferentGroupDoesntWork(self, connection, group, capsys):
+    def testReuseFromDifferentGroupDoesntWork(self, connection, group):
         """
         Test session reuse with different groups fails
         """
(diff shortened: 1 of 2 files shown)


sbesson:
As suggested by 606f9fc, one downside of failing in `sessions login` when `stdout` is not a `tty` is tests can no longer simply use [pytest capsys](https://docs.pytest.org/en/stable/capture.html) fixture to test scenarios where additional input to the CLI control is required. I expect most (if not all) integration tests should be unaffected by this change.

dominikl:
That's good, thanks Seb 👍 Often I've been running `omero something >> out.log` just to discover later that it's been sitting there and waiting for credentials.

test/unit/clitest/test_sess.py
```diff
@@ -423,9 +425,6 @@ def testReuseFromDifferentGroupDoesntWork(self, connection, group, capsys):
         conn_args = self.get_conn_args(connection, group=group[1])
         cli.invoke(["s", "login"] + conn_args)
         cli.assertReqSize(self, 0)
-        out, err = capsys.readouterr()
-        msg = (self.get_conflict_message() + 'omero.group: %s!=%s')
-        assert err.splitlines()[-2] == msg % ('testsessid', group[0], group[1])
```
manics, line 428: Was this testing anything of particular importance?

manics (approved):
LGTM. One question about the removed test.

test/unit/clitest/test_sess.py
sbesson: This was primarily testing the error message thrown when trying to log in under a different group with an existing session. As mentioned in https://github.com/ome/omero-py/pull/256#issuecomment-700934845, the new stdout/tty requirements makes this incompatible with testing the output via `capsys` in that case.

The primary logic of the test i.e. the fact that some user input is required is still tested so I don't believe we are losing much. I can capture this as a RFE issue to resurrect and expand the sessions error message testing.

sbesson:
Should I go ahead and start the process of releasing this as `omero-py 5.8.2` or is anything else planned?

manics:
I'm not aware of anything, maybe check in standup tomorrow?

PR #257, opened 2020-09-29
Remove f-string usage for Python 3.5 compatibility

sbesson:
The usage of f-strings breaks Python 3.5 compatibility which affects the ability to run OMERO on some environments. This is a minor change which restores some compatibility.


joshmoore (approved):
Looks good to me.

joshmoore:
(I also don't see other hits trying various regexes like 
```
  492  git grep -E '[^"-a-z]f".*"'
  496  git grep -E "[^'-]f'.*'"
```
)

PR #261, opened 2020-10-01
Add GitHub workflow

manics:
I got fed up of waiting for Travis. If we decide to go this route we'll need to setup (and test) the Pypi deploy secrets, and should delete `.travis.yml`

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,0 +1,62 @@
+# https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Build
+on:
+  push:
+  pull_request:
+
+jobs:
+
+  test:
+    name: Test
+    strategy:
+      # Keep running so we can see if other tests pass
+      fail-fast: false
+      matrix:
+        python-version:
+          - '3.6'
+          - '3.7'
+          - '3.8'
+        os:
+          - ubuntu-20.04
+        include:
+          - python-version: '3.7'
+            os: macos-latest
+          - python-version: '3.7'
+            os: windows-latest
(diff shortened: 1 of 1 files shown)


sbesson:
I am getting increasingly convinced of our upcoming transition to GitHub workflows so no objection from my side. A few questions:

- https://github.com/ome/omero-py/pull/261/checks is not displaying anything. Does something need to be activated?
- should we try a PyPI release via GitHub action for `omero-py 5.8.2`?

joshmoore:
> * Does something need to be activated?

I'm not convinced adding a workflow in a PR always works as expected (or at all?)


> * should we try a PyPI release via GitHub action for `omero-py 5.8.2`?

WFM.

manics:
> I'm not convinced adding a workflow in a PR always works as expected (or at all?)

It usually works for me, but maybe it depends on the repo settings?

sbesson:
I am tempted in that case to believe https://github.com/manics/omero-py/commit/c7c1074e6595ca85facef437ed67217ca68b0589 and merge. Although we probably need to remove at least the Travis deployment step to avoid having 2 places releasing concurrently. Separate PR?

manics:
I've removed the deploy step so we can merge it and test the workflow runs as expected. I'll then add the deploy step back and remove travis in the next PR

sbesson:
:+1: a general question I have is about what the name of GitHub workflow should be esp. if there is a single one. Options I can think of `workflow.yml`, `main.yml`, `build.yml`. This does not need ot block this PR for the reasons described above.

sbesson:
See https://github.com/ome/omero-py/actions/runs/282610505

PR #274, opened 2021-01-22
download provider uses NamedTemporaryFile

will-moore:
See https://github.com/ome/omero-scripts/pull/180

The Populate_Metadata script that uses `DownloadingOriginalFileProvider` needs to read the temp file with a specific encoder. By returning a `NamedTemporaryFile`, consumers can open the temp file themselves.


dominikl:
👍 Makes sense, good to merge. I would never have expected that `tempfile.TemporaryFile()` can create a file without a "visible" file name, Python's always good for a surprise.

PR #276, opened 2021-01-27
createImageFromNumpySeq handles big images

will-moore:
This adds support for big images to `conn.createImageFromNumpySeq()`.

Previously this failed if the numpy planes were larger than 3k * 3k.

Fixes https://github.com/ome/omero-figure/issues/403

I was hoping that this method would "just work" exactly the same with big images and small images, so you didn't have to worry about whether you were creating a big image or not.
Unfortunately, the `TileLoop` for populating big images with `setTile()` wants to consume planes in the order:

```
for t in range(size_t):
    for c in range(size_c):
       for z in range(size_z):
           ...
```
so the `planeGenerator` passed to ``conn.createImageFromNumpySeq()` needs to provide planes in that order.
But for small images, we need:

```
for z in range(size_z):
    for c in range(size_c):
       for t in range(size_t):
           ...
```

Obviously this is not ideal. But I couldn't think of any way to populate Big Images with the order of planes that is already required by `conn.createImageFromNumpySeq()`.

NB: the OMERO.figure issue above is still fixed, since we only have a single Z and T :)

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -3731,13 +3731,18 @@ class _BlitzGateway (object):
         return self.countAnnotations(obj_type, obj_ids)
 
 
-    def upload_tiles(self, image, zct_plane_gen):
+    def upload_tiles(self, image, zct_plane_gen, tile_size=256):
+        """
+        Upload numpy planes to a 'big' image as tiles.
+
+        @param  image           omero.model.ImageI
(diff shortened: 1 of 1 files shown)


will-moore:
Closing since this isn't ready and it's now conflicting with https://github.com/ome/omero-py/pull/290#issuecomment-851731411

PR #279, opened 2021-02-23
Bug: `obj update` strips newlines

joshmoore:
The regex used for capturing the string after the
= symbol currently stops at the first newline. This
passes `re` flags to allow capturing the multiline.

More tests are needed.

see: https://forum.image.sc/t/line-breaks-in-descriptions-via-cli/49236/4

### DIFF
--- a/src/omero/plugins/obj.py
+++ b/src/omero/plugins/obj.py
@@ -43,7 +43,7 @@ class TxField(object):
 
     ARG_RE = re.compile((r"(?P<FIELD>[a-zA-Z][a-zA-Z0-9]*)"
                          "(?P<OPER>[@])?="
-                         "(?P<VALUE>.*)"))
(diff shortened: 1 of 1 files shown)


sbesson:
Tested the two commands suggested in https://github.com/ome/openmicroscopy/pull/6265 against a CI server with this fix included. Everything works as expected and multi-line descriptions render as multi-line.

The `omero obj update Dataset:123 description=$'A\nB'` syntax is not very intuitive so I don't know whether this should be captured in the help or whether the expectation is simply that someone trying to update a field with the CLI will naturally use the second form.

Otherwise LGTM

sbesson (approved):

joshmoore:
Yeah, it'll be dependent on someone's shell, etc etc.

will-moore:
Tested locally. 👍

PR #280, opened 2021-03-03
config add web middleware

will-moore:
See https://github.com/ome/omero-web/pull/260

With an update to the default `omero.web.middleware` as in the PR above,
if the user has previously updated their config then the values will be stored in the `etc/grid/config.xml` and
the new default will be ignored.
This PR aims to update the `omero.web.middleware` config in the same way as the above PR.
This will only happen if the above PR is included in `omero-web` (the new middleware is importable)

NB: last commit of the PR above changes the default `omero.web.middleware` setting so that it is simpler
for this PR to update in the same way (simply append a single class, instead of update multiple classes in the list).
This allows us to simplify the update and reduce the amount of settings that are hard-coded in `config.py`.

I'm assuming no-one is upgrading from "4.2.1" (6 years ago).

To test:

 - with current web (without https://github.com/ome/omero-web/pull/260) do typical CORs config to update middleware:
 ```
$ omero config append omero.web.middleware '{"index": 0.5, "class": "corsheaders.middleware.CorsMiddleware"}'
$ omero config append omero.web.middleware '{"index": 10, "class": "corsheaders.middleware.CorsPostCsrfMiddleware"}'
```
 - Update `omero-py` to include this PR
 - The config should not be changed yet (this won't show `CustomHeadersMiddleware`
```
$ omero config get omero.web.middleware
```
- Update omero-web to include https://github.com/ome/omero-web/pull/260
- The config *should* now show `CustomHeadersMiddleware`
```
$ omero config get omero.web.middleware
```

NB: If you *ONLY* stop web, upgrade web and start web (without any `$ omero config... ` command), the change will not take effect.

### DIFF
--- a/src/omero/config.py
+++ b/src/omero/config.py
@@ -227,16 +227,29 @@ class ConfigXml(object):
         for k, v in self.properties(None, True):
             version = self.version(k)
             if version == "5.1.0" and v is not None:
-                for x in list(v):
-                    # User has configured their middleware list.
-                    if x.get("name") == "omero.web.middleware":
-                        val = x.get("value", "")
-                        middleware = json.loads(val)
-                        middleware.append({"index": 7, "class": "omeroweb.middleware.CustomHeadersMiddleware"})
-                        val = json.dumps(middleware)
-                        x.set("value", val)
-                    if x.get("name") == self.KEY:
-                        x.set("value", self.VERSION)
+                try:
+                    from omeroweb.middleware import CustomHeadersMiddleware
+                except:
+                    self.logger.info(
+                        "Failed to import"
+                        " omeroweb.middleware.CustomHeadersMiddleware")
+                else:
+                    self.logger.info(
+                        "Adding omeroweb.middleware.CustomHeadersMiddleware"
+                        " to omero.web.middleware")
+                    for x in list(v):
+                        # User has configured their middleware list.
+                        if x.get("name") == "omero.web.middleware":
+                            val = x.get("value", "")
+                            middleware = json.loads(val)
+                            middleware.append({
+                                "index": 7,
+                                "class":
+                                "omeroweb.middleware.CustomHeadersMiddleware"})
+                            val = json.dumps(middleware)
+                            x.set("value", val)
+                        if x.get("name") == self.KEY:
+                            x.set("value", self.VERSION)
 
     def version_fix(self, props, version):
         """
--- a/src/omero/config.py
+++ b/src/omero/config.py
@@ -238,7 +238,7 @@ class ConfigXml(object):
                 try:
                     from omeroweb.middleware import CustomHeadersMiddleware
                 except:
-                    self.logger.info(
+                    self.logger.error(
                         "Failed to import"
                         " omeroweb.middleware.CustomHeadersMiddleware."
                         " Please upgrade omero-web")
--- a/src/omero/config.py
+++ b/src/omero/config.py
@@ -244,8 +244,8 @@ class ConfigXml(object):
                         " Please upgrade omero-web")
                 else:
                     self.logger.info(
-                        "Adding omeroweb.middleware.CustomHeadersMiddleware"
-                        " to omero.web.middleware")
+                        "Removing default middleware classes"
+                        " from omero.web.middleware")
                     for x in list(v):
                         # User has configured their middleware list.
                         if x.get("name") == "omero.web.middleware":
--- a/src/omero/config.py
+++ b/src/omero/config.py
@@ -232,7 +232,8 @@ class ConfigXml(object):
                 except:
                     self.logger.info(
                         "Failed to import"
-                        " omeroweb.middleware.CustomHeadersMiddleware")
+                        " omeroweb.middleware.CustomHeadersMiddleware."
+                        " Please upgrade omero-web")
                 else:
                     self.logger.info(
                         "Adding omeroweb.middleware.CustomHeadersMiddleware"
@@ -240,7 +241,7 @@ class ConfigXml(object):
                     for x in list(v):
                         # User has configured their middleware list.
                         if x.get("name") == "omero.web.middleware":
-                            val = x.get("value", "")
+                            val = x.get("value", "[]")
                             middleware = json.loads(val)
                             middleware.append({
                                 "index": 7,
--- a/src/omero/config.py
+++ b/src/omero/config.py
@@ -87,7 +87,7 @@ class ConfigXml(object):
     in etc/grid. For a copy of the dict, use "as_map"
     """
     KEY = "omero.config.version"
-    VERSION = "5.1.0"
+    VERSION = "5.2.0"
     INTERNAL = "__ACTIVE__"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/omero/config.py [hunk 1]
```diff
@@ -222,6 +223,34 @@ def toplinks_check(self):
...
 
+    def middleware_check(self):
+        for k, v in self.properties(None, True):
+            version = self.version(k)
+            if version == "5.1.0" and v is not None:
+                try:
+                    from omeroweb.middleware import CustomHeadersMiddleware
+                except:
+                    self.logger.info(
+                        "Failed to import"
```
joshmoore, line 234: Does this mean the user should upgrade omero-web? Worth saying so?

src/omero/config.py
```diff
@@ -222,6 +223,34 @@ def toplinks_check(self):
                     if x.get("name") == self.KEY:
                         x.set("value", self.VERSION)
 
+    def middleware_check(self):
+        for k, v in self.properties(None, True):
+            version = self.version(k)
+            if version == "5.1.0" and v is not None:
```
joshmoore, line 229: This block is only really needed `if x == omero.web.middleware`, right?

src/omero/config.py
```diff
@@ -222,6 +223,34 @@ def toplinks_check(self):
...
+                        "Failed to import"
+                        " omeroweb.middleware.CustomHeadersMiddleware")
+                else:
+                    self.logger.info(
+                        "Adding omeroweb.middleware.CustomHeadersMiddleware"
+                        " to omero.web.middleware")
+                    for x in list(v):
+                        # User has configured their middleware list.
+                        if x.get("name") == "omero.web.middleware":
+                            val = x.get("value", "")
```
joshmoore, line 243: Should the default be `[]` or `{}`? Doesn't `json.loads("")` throw?

joshmoore (changes requested):
A few suggestions for modification. Additionally, this solves the specific issue, but doesn't solve the overall design bug. That's certainly a trade-off we can make, but just to be clear.

src/omero/config.py
will-moore: We don't have access to `x` at this point. `v` is a list of properties and we need to update the value of `omero.web.middleware` and the `omero.config.version`.

src/omero/config.py
will-moore: fixed in d4e16be

src/omero/config.py
will-moore: fixed in d4e16be

src/omero/config.py
joshmoore: Sure, but the for loop over v could be moved up to this point, no?

joshmoore (dismissed):

will-moore:
@joshmoore I could maybe do a better fix by creating a `INSTALLED_MIDDLEWARE` and `ADDITIONAL_MIDDLEWARE` in https://github.com/ome/omero-web/pull/260 (just as we have for  `omero.web.apps` with `INSTALLED_APPS` and `ADDITIONAL_APPS`). Then use this `config.py` update to only leave user-configured middleware in the `omero.web.middleware` (ADDITIONAL_MIDDLEWARE) list.
I might give it a try and see how it looks...
Don't know if/when we might want to apply this to other configs. There's not many that we are likely to update.

joshmoore:
> Don't know if/when we might want to apply this to other configs.

Are there others where we are copying a default?

will-moore:
I could list them. Looking for where we use `json.loads` and the default isn't empty:

```
"omero.web.caches": '{"default": {"BACKEND":' ' "django.core.cache.backends.dummy.DummyCache"}}'
"omero.web.server_list": '[["%s", 4064, "omero"]]' % CUSTOM_HOST
"omero.web.open_with": '[["Image viewer", "webgateway", {"supported_objects": ["image"],"script_url": "webclient/javascript/ome.openwith_viewer.js"}]]'
"omero.web.ui.top_links": (Data, History, Help)
"omero.web.ui.metadata_panes": (tag, map, table, file, comment, rating, other)
"omero.web.ui.right_plugins":  (Aquisition, Preview)
```

I don't *expect* many of those defaults to change, or for users to have them configured. Except maybe `top_links` and `open_with`. BUT, if we only allow users to ADD to the defaults, then they can't remove items from defaults. E.g. remove `metadata_panes`.
The same applies to `omero.web.middleware`. If someone has already removed an item from the default settings, then putting all the defaults into a non-editable list would add it back. So we can't support `remove` of defaults AND auto update of defaults.
I think I'll try the separation of `omero.web.middleware` into 'installed' and 'additional' lists, and we can see what that looks like. Then we can decide if we want to do the same with `open_with` and/or `top_links`.

will-moore:
With https://github.com/ome/omero-web/pull/260/commits/70d22293e45a76d2fd47ab56f48bc41f61688719 and the last commits above, the user settings are now distinct from the default middleware.

src/omero/config.py
joshmoore: This only gets printed if the user sets `DEBUG=1 ...` which seems pretty unlikely.

src/omero/config.py
```diff
@@ -222,6 +223,41 @@ def toplinks_check(self):
                     if x.get("name") == self.KEY:
                         x.set("value", self.VERSION)
 
+    def middleware_check(self):
+        mw_classes = [
+            "django.middleware.common.BrokenLinkEmailsMiddleware",
```
joshmoore, line 228: So there's no way to disable this middleware, right? Are there any cases you can think of where that would be necessary?

joshmoore (commented):
Testing locally:
 * In a new environment, ran `append` and the full list was cached.
 * Updated omero-py and ran get, the old version was shown.
 * Ran with `DEBUG=1` and could see the message telling me to upgrade.
 * Updated omero-web and was only shown the properties I had set.

Questions inline.

joshmoore:
@sbesson was kind enough to scour the history to find https://github.com/ome/openmicroscopy/pull/2030#issuecomment-33578012 -- pasting it here for reference.

src/omero/config.py
will-moore: I'm sorry, I'm not seeing it. The loop over v starts right after I check if `v is not None`.

src/omero/config.py
will-moore: No, I can't think of any.

will-moore:
I don't think we're going to be able to apply the same approach to the other settings lists above.
For the UI settings (and open_with), it's possible a user will have configured them to remove one or more default items.
For `web.caches`, this isn't a list, so if it's been configured it will likely be `set` rather than `append` to default.
I guess for `server_list`, we might want to change the one default option at some point. But I think we could leave any upgrade path in config.py till it's needed.

src/omero/config.py
joshmoore: No worries. It's a super minor optimization and this isn't critical code.

will-moore:
With https://github.com/ome/omero-web/pull/260 closed for now, don't need this currently. Closing...

PR #285, opened 2021-04-06
get objects by map annotations

will-moore:
See request for "How to query objects by map annotation?"
We don't have any examples at https://docs.openmicroscopy.org/omero/5.6.3/developers/Python.html
and need to look into Mapr source to find out how.
https://forum.image.sc/t/harmonization-of-image-metadata-for-different-file-formats-omero-mde/50827/10

This adds method  to gateway:

```
conn.getObjectsByMapAnnotations('Image', key='Primary Antibody')

conn.getObjectsByMapAnnotations('Image', value='ACA', ns="openmicroscopy.org/omero/client/mapAnnotation")

conn.getObjectsByMapAnnotations('Image', key='Primary Antibody', value='ACA')
```

To consider:

 - We could allow wild-card queries (as used for auto-complete in mapr)
 - One way to support this would be to allow the 'key' or 'value' to start/end with `*`: e.g. `key='Anti*' `
 - The BlitzGateway code could replace the `*` with `%%` and use `like` in the HQL query.

TODO: 
 - add tests... (Done - thanks Dom)
 - Add docstring

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -4183,6 +4183,45 @@ class _BlitzGateway (object):
         for e in q.findAllByQuery(sql, p, self.SERVICE_OPTS):
             yield wrapper(self, e)
 
+    def getObjectsByMapAnnotations(self, obj_type, key=None, value=None, ns=None,
+                                   opts={}):
+
+        wrapper = KNOWN_WRAPPERS.get(obj_type.lower(), None)
+        if not wrapper:
+            raise AttributeError("Don't know how to handle '%s'" % obj_type)
+
+        params = omero.sys.ParametersI()
+        clauses = []
+        if key is not None:
+            clauses.append("mv.name = :key")
+            params.addString("key", key)
+        if value is not None:
+            clauses.append("mv.value = :value")
+            params.addString("value", value)
+        if ns is not None:
+            clauses.append("ann.ns = :ns")
+            params.addString("ns", ns)
+
+        # Parse opts dict to build params
+        limit = opts.get('limit', 500)
+        offset = opts.get('offset', 0)
+        if offset is not None and limit is not None:
+            params.page(offset, limit)
+
+        # Using projection since can't seem to fetch objects AND filter by mapValue
+        query = """
+            select obj.id, obj.name from
+            %sAnnotationLink ial
+            join ial.child ann
+            join ann.mapValue mv
+            join ial.parent obj""" % wrapper().OMERO_CLASS
+        if len(clauses) > 0:
+            query += " where " + " and ".join(clauses)
+        result = self.getQueryService().projection(query, params, self.SERVICE_OPTS)
+        ids = [row[0].val for row in result]
+        return self.getObjects(obj_type, ids)
+
+
     ################
     # Enumerations #
 
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -4192,12 +4192,24 @@ class _BlitzGateway (object):
 
         params = omero.sys.ParametersI()
         clauses = []
-        if key is not None:
-            clauses.append("mv.name = :key")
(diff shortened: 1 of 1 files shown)


pwalczysko:
```
...
result2 = conn.getObjectsByMapAnnotations('Image', value='bla ', ns="openmicroscopy.org/omero/client/mapAnnotation")
	print ("result2", "\n")
	print(result2)
	for i in result2:
		print(i.name, "\n", i.id)
...
```

The above bit of code is inside my script. When executed, it gives

```
serverStackTrace = ome.conditions.ApiUsageException: unexpected end of subtree [select obj from ome.model.core.Image obj join fetch obj.details.owner as owner join fetch obj.details.creationEvent where obj.id in ()]; nested exception is org.hibernate.hql.ast.QuerySyntaxException: unexpected end of subtree [select obj from ome.model.core.Image obj join fetch obj.details.owner as owner join fetch obj.details.creationEvent where obj.id in ()]
```

This is caused purely by having the value as 
``value='bla '`` # note the indtentional whitespace after ``bla``

When no whitespace is inserted, then the method does not crash. Considering the issues described in https://forum.image.sc/t/harmonization-of-image-metadata-for-different-file-formats-omero-mde/50827/14, this behaviour should be fixed imho.

pwalczysko:
used folllowing workflow to set up the testing env:

1. create a new conda env as per https://docs.openmicroscopy.org/omero/5.6.3/users/cli/installation.html
2. run following:
```
pip uninstall omero-py
pip install git+https://github.com/will-moore/omero-py.git@get_objects_by_map_annotations
```

Use a script which contains lines from the description of this PR

pwalczysko:
What is the way to sanitize one's script against the case that no images with specified KVPs were found ? The "generator" object which is returned by the methods added here has not a length zero even if turned into a list and checked for length ?

pwalczysko:
Just tried to read the retrieved Images (objects) by the methods added in this PR just as specified in https://docs.openmicroscopy.org/omero/5.6.3/developers/Python.html#read-data

But, again, when no images are found, the same error as in https://github.com/ome/omero-py/pull/285#issuecomment-814987008 happened. Maybe something is really wrong with the Image objects (generators) returned by the new methods here ?

src/omero/gateway/__init__.py
```diff
@@ -4183,6 +4183,45 @@ def getObjectsByAnnotations(self, obj_type, annids):
...
+        # Using projection since can't seem to fetch objects AND filter by mapValue
+        query = """
+            select obj.id, obj.name from
+            %sAnnotationLink ial
+            join ial.child ann
+            join ann.mapValue mv
+            join ial.parent obj""" % wrapper().OMERO_CLASS
+        if len(clauses) > 0:
+            query += " where " + " and ".join(clauses)
+        result = self.getQueryService().projection(query, params, self.SERVICE_OPTS)
```
dominikl, line 4220: I can confirm @pwalczysko error, it happens when there are no images found. I think some check has to happen here if `self.getQueryService().projection` has actually returned something.

pwalczysko:
Tested all three options in the header of this PR.  All works, including the cases where no images are retrieved. Also tried to add the KVP on a Project, worked too. All good fmpov.

src/omero/gateway/__init__.py
```diff
@@ -4183,6 +4183,59 @@ def getObjectsByAnnotations(self, obj_type, annids):
         for e in q.findAllByQuery(sql, p, self.SERVICE_OPTS):
             yield wrapper(self, e)
 
+    def getObjectsByMapAnnotations(self, obj_type, key=None, value=None, ns=None,
+                                   opts={}):
```
will-moore, line 4187: TODO: Need to add docstring!

joshmoore:
In general, probably a good example of something that needs to be moved server-side (or into the eventual JSON API).

:+1:

PR #287, opened 2021-05-18
Add logic to search OMERO.tables on non Pythonic named columns

sbesson:
See https://github.com/ome/omero-metadata/issues/57 for the reference issue.

OMERO.tables HDF storage uses PyTables [querying capabilities](https://www.pytables.org/usersguide/libref/structured_storage.html#table-methods-querying) in particular the `tables.get_where_list` API. This API takes as a primary input a condition which syntax is defined as in https://www.pytables.org/usersguide/condition_syntax.html.

Columns named in a Python compliant way (i.e. matching `^[a-zA-Z_][a-zA-Z0-9_]*$`), the column name can directly be used in the condition e.g. `foo>1`. Column names that do not match this patterns e.g. including spaces or brackets are supported but they need to be retrieved using `getattr` and they cannot directly used as an input condition. As it stands, this prevents the API from querying all the row based on a condition including such a column.

350dbc1 demonstrates how to use `condvars` and variable substitution to allow searching these non Pythonic named columns. ae5ca40 proposes to expand the semantics of the `variables` argument of `tables.get_where_list` to support these queries remotely.  Any string/string key value pair is expanded as a string/column key/pair and passed to the underlying `get_where_list` API allowing variable substitution. 

Using the OMERO API, this should allow the following types of queries

```
table.getWhereList("(count > 10) & (x =='pass')", variables={'x': rstring('Quality Control')}, start=0, stop=rowCount, step=0)
```

### DIFF
--- a/src/omero/hdfstorageV2.py
+++ b/src/omero/hdfstorageV2.py
@@ -543,7 +543,12 @@ class HdfStorage(object):
                      start, stop, step):
         self.__initcheck()
         try:
-            return self.__mea.get_where_list(condition, variables, None,
+            condvars = variables
+            if variables:
+                for key, value in condvars.items():
+                    if isinstance(value, str):
+                        condvars[key] = getattr(self.__mea.cols, value)
+            return self.__mea.get_where_list(condition, condvars, None,
                                              start, stop, step).tolist()
         except (NameError, SyntaxError, TypeError, ValueError) as err:
             aue = omero.ApiUsageException()
--- a/test/unit/tablestest/test_hdfstorage.py
+++ b/test/unit/tablestest/test_hdfstorage.py
@@ -304,6 +304,28 @@ class TestHdfStorage(TestCase):
         # Doesn't work yet.
         hdf.cleanup()
 
+    @pytest.mark.parametrize('name', [
+        "my name", "name [unit]", "name (unit)"])
+    def testNonNaturalNameWarning(self, name):
+        # Tests String column with non-pythonic names
+        # See also Pytables NaturalNameWarning
+        bytesize = 3
+        hdf = HdfStorage(self.hdfpath(), self.lock)
+        cols = [omero.columns.StringColumnI(
+            name, "A column with non natural name", bytesize, None)]
+        hdf.initialize(cols)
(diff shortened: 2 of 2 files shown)


sbesson:
--exclude to review https://github.com/ome/omero-py/pull/288

jburel:
```
table.getWhereList("(count > 10) & (x =='pass')", variables={'x': 'Quality Control'}, start=0, stop=rowCount, step=0)
```
This is not what the test does ``rstring("Long column")`` cf. https://github.com/ome/openmicroscopy/pull/6283
I think it is preferable not to ask user of the API to have to use ``rstring``

sbesson:
PR description updated although the primary issue is rather at the level of the `omero-blitz` component as the [Table.getWhereList](https://docs.openmicroscopy.org/omero-blitz/5.5.8/slice2html/omero/grid/Table.html#getWhereList) API expects `variables` to be of type [RTypeDict](https://docs.openmicroscopy.org/omero-blitz/5.5.8/slice2html/omero.html#RTypeDict). 

I would also be of the opinion of lowering the exposition of these types for consumers of the API if possible. What are the viable options?

joshmoore:
> What are the viable options?

A gateway method would be able to add a call to `omero.rtypes.wrap`, but we don't currently have a general interceptor for direct calls to Prx instances.

joshmoore:
Thoughts on getting this out with #292?

sbesson:
Answering https://github.com/ome/omero-py/pull/287#issuecomment-877390930 specifically, from a code perspective,  I think the change here should be fairly harmless and releasable immediately. The[ PyTables API expectation](https://www.pytables.org/usersguide/libref/structured_storage.html#tables.Table.where) is that the `condvars` variable will contain a dictionary pointing to `Column` instances. I have not been able to think of a  scenario where the special handling of string values implemented here would break existing behaviors.

As discussed earlier this morning, I do not know of any immediate need for this fix so it probably makes sense to allow more time for the review of this PR alongside the examples and some technical documentation explaining its impact for API consumers.

joshmoore (approved):
In general, lgtm. I haven't been able to come with a scenario where I think existing code would be unhappy (and I also don't think there's too much of that)

PR #288, opened 2021-05-27
Fix HdfStorage.read() implementation to work with all column types

sbesson:
Reported by @jburel, the tables.read() method internally uses `col.values` (via `_getrows`). For any column type with overrides the default methods of `AbstractColumn` like `MaskColumnI`, this will fail.

fe8b863 tried to address the problem by using the `getsize()` API which should be independent of the column implementation details. However, `testMaskColumn` is still failing as the logic does not populate the `.bytes` attributes.

 6c27eae is a more robust fix to the underlying issue which uses the fact that the column API effectively has a `read` method and effectively delegates to this API like `readCoordinates`.

All the other commits expand the unit tests to cover both `readCoordinates` and `read` scenarios for the different column types (long, string, mask) as well as a test for various arguments of `table.read`

Proposing for a patch release of `omero-py`

### DIFF
--- a/src/omero/hdfstorageV2.py
+++ b/src/omero/hdfstorageV2.py
@@ -590,20 +590,6 @@ class HdfStorage(object):
 
         return self._as_data(cols, rowNumbers)
 
-    def _getrows(self, start, stop):
-        return self.__mea.read(start, stop)
-
-    def _rowstocols(self, rows, colNumbers, cols):
-        l = 0
-        rv = []
-        for i in colNumbers:
-            col = cols[i]
-            col.fromrows(rows)
-            rv.append(col)
-            if not l:
-                l = col.getsize()
-        return rv, l
-
     @stamped
     def slice(self, stamp, colNumbers, rowNumbers, current):
         self.__initcheck()
--- a/src/omero/hdfstorageV2.py
+++ b/src/omero/hdfstorageV2.py
@@ -581,7 +581,14 @@ class HdfStorage(object):
 
         for col in cols:
             col.read(self.__mea, start, stop)
-        return self._as_data(cols, range(stop - start + 1))
+        if start is not None and stop is not None:
+            rowNumbers = list(range(start, stop))
+        elif start is not None and stop is None:
+            rowNumbers =  list(range(start, start + 1))
+        elif start is None and stop is None:
+            rowNumbers = list(range(self.__length()))
+
+        return self._as_data(cols, rowNumbers)
 
     def _getrows(self, start, stop):
         return self.__mea.read(start, stop)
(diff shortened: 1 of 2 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#707](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/707/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/707/consoleText) for more details.
Possible conflicts:
  - PR #287 sbesson 'Add logic to search OMERO.tables on non Pythonic named columns'
    - test/unit/tablestest/test_hdfstorage.py

--conflicts

src/omero/hdfstorageV2.py
```diff
@@ -576,11 +576,12 @@ def readCoordinates(self, stamp, rowNumbers, current):
     def read(self, stamp, colNumbers, start, stop, current):
         self.__initcheck()
         self.__sizecheck(colNumbers, None)
-        cols = self.cols(None, current)
+        all_cols = self.cols(None, current)
+        cols = [all_cols[i] for i in colNumbers]
 
-        rows = self._getrows(start, stop)
```
joshmoore, line 581: The `_getrows()` and `_rowstocols()` definitions can be removed now, no?

sbesson: Most likely, I would need to review their usage in other APIs which I haven't reviewed. They are marked as internal so happy to remove or deprecate.

src/omero/hdfstorageV2.py
```diff
@@ -576,25 +576,19 @@ def readCoordinates(self, stamp, rowNumbers, current):
...
-        cols = self.cols(None, current)
-
-        rows = self._getrows(start, stop)
-        rv, l = self._rowstocols(rows, colNumbers, cols)
-        return self._as_data(rv, list(range(start, start + l)))
+        all_cols = self.cols(None, current)
+        cols = [all_cols[i] for i in colNumbers]
 
-    def _getrows(self, start, stop):
-        return self.__mea.read(start, stop)
```
joshmoore, line 586: The only thing I'd have to check is if there are any situations where the single read is significantly faster, though I assume it's a time v. space issue.

joshmoore (approved):
One minor item I noticed which we might want to just consider later, otherwise :+1:

src/omero/hdfstorageV2.py
sbesson: I certainly see how the performance can vary depending of the ratio percentage_of_read_columns/percentage_of_read_rows.

I would assume `readCoordinates` will suffer from the same tradeoffs as it it also looping over the individual columns to extract records and combine them into a grid.

sbesson: Objections to capture this as an issue and merging

joshmoore: WFM

PR #297, opened 2021-07-23
point to image.sc

jburel:
Point to image.sc
cc @dominikl

### DIFF
--- a/src/omero/plugins/basics.py
+++ b/src/omero/plugins/basics.py
@@ -169,7 +169,7 @@ Other help topics:
 
 For additional information, see:
 https://docs.openmicroscopy.org/latest/omero/users/cli/index.html
-Report bugs at https://forum.image.sc/
+Report bugs at https://forum.image.sc/tag/omero-cli
 """
 
 
--- a/src/omero/plugins/basics.py
+++ b/src/omero/plugins/basics.py
@@ -169,7 +169,7 @@ Other help topics:
 
 For additional information, see:
 https://docs.openmicroscopy.org/latest/omero/users/cli/index.html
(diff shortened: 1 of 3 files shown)


dominikl:
Thanks, looks good 👍

src/omero/plugins/basics.py
```diff
@@ -169,7 +169,7 @@ def __call__(self, args):
 
 For additional information, see:
 https://docs.openmicroscopy.org/latest/omero/users/cli/index.html
-Report bugs to <ome-users@lists.openmicroscopy.org.uk>
+Report bugs at https://forum.image.sc/
```
sbesson, line 172: An alternate link since we are talking about bugs would be https://forum.image.sc/c/usage-issues/7 but I am not sold on it and pointing to the top-level URL might be the better choice here.

sbesson (approved):

src/omero/plugins/basics.py
joshmoore: Since we monitor tags rather than categories, I tend to default to the likes of https://forum.image.sc/tag/omero-insight and then we can change the category ourselves if need be.

joshmoore: If there are no objections, I'll push a commit here to use `tag/omero-cli` and then @dominikl can move forward with https://github.com/ome/omero-documentation/pull/2192

sbesson: @joshmoore ping. Are you still pushing an extra commit?

joshmoore: Apparently I wasn't but now I have.

sbesson: :+1:

PR #304, opened 2021-09-08
Trigger pr

jburel:
This PRs opens a PR against https://github.com/ome/conda-omero-py when a new version of the package is published on https://pypi.org/

snoopy will create the PR and needs to be a collaborator on the target repo i.e. https://github.com/ome/conda-omero-py


.github/workflows/publish.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          password: ${{ secrets.PYPI_PASSWORD }}
+  update-repository:
+    needs:
+      - publish-pypi
+    runs-on: ubuntu-latest
+    steps:
+      - name: Extract version and sha256
+        run: |
+          python -m pip download --no-deps --no-binary :all: omero-py
```
joshmoore, line 34: I wonder if these steps are something that `scc` or similar should learn.

sbesson: Also if we are thinking of a generic tool, we should be able to sidestep the download completely and rely on the PyPI JSON API e.g. https://pypi.org/pypi/omero-py/json

jburel: I will check what I can do with the json in shell script.
Tool like jq might be useful

jburel: parsing the json  might not simplify things
```
json=$(curl -s 'https://pypi.org/pypi/omero-py/json')
value=$(echo ${json} | python -mjson.tool | grep "sha256" | tail -1)
sha256="${value#*sha256\": \"}"
sha256="${sha256%\"}"
echo "${sha256}"
name=$(echo ${json} | python -mjson.tool | grep ".tar.gz" | tail -1)
version="${name#*omero-py-}"
version="${version%.tar.gz\",}"
echo "${version}"
```

jburel: Using jq and getting the version from the github tag, we could have
```
while IFS='' read release;do
    for name in $(echo "$release" | jq -r '.filename | endswith(".tar.gz") '); do
      if $name; then
        sha256=$(echo "$release" | jq -r '.digests.sha256')
     fi
    done
done <<< "$(echo ${json} | jq -c --arg tag "${tag}" '.releases[$tag][]')"

```

jburel: The version using ``jq`` is more robust since it does not assume the order in the json

joshmoore: No objection to `jq` but I think I'm more concerned about the maintainability of the total lines in workflow yaml than whether it's bash or not.

jburel: I don't think there is an easy way around it.

joshmoore: Not even with: `pip install $TOOL` or `wget $URL/tool.sh`?

jburel: but you still need to extract the sha256 or calculate it as in the first option

joshmoore: Guess I see all of these lines:

```
          python -m pip download --no-deps --no-binary :all: omero-py
          output=$(python -m pip hash *.gz)
          hash256=$(echo $output | cut -d':' -f 3)
          package=$(echo $output | cut -d':' -f 1)
          version="${package#*omero-py-}"
          version="${version%.tar.gz}"
          echo "package_sha256="$hash256 >> $GITHUB_ENV
          echo "package_version="$version >> $GITHUB_ENV
          rm *.gz
          sed -i "s/\(.*sha256:.*\)/sha256: ${{env.package_sha256}}/g" meta.yaml
          sed -i "s/\(version = \).*/version = \"${{env.package_version}}\" %}/" meta.yaml
          git config user.name 'snoopycrimecop'
          git config user.email 'snoopycrimecop@users.noreply.github.com'
          git add -A
          git commit -m "Update version"
```
as being:

```
scc bump-conda-version $VERSION
```

jburel: you mean ``scc bump-conda-version $VERSION $SHA``

joshmoore: > you mean scc bump-conda-version $VERSION $SHA

Can't the tool/script calculate the SHA as well? Maybe it's just:

```
scc bump-conda-version omero-py
```
?

jburel: sha256 
will be retrieve from various repositories e.g.
https://github.com/glencoesoftware/raw2ometiff/releases/download/v{{ version }}/raw2ometiff-{{ version }}.zip
another one
https://downloads.openmicroscopy.org/bio-formats/{{ PKG_VERSION }}/artifacts/bftools.zip

Basically each "conda" repo has its own logic.

jburel:
Closing. The logic has been implemented in scc

Issue #306, opened 2021-10-11
Error with omero tag list when a lot of tags


lldelisle:
Hi,
I just installed omero with conda using the documentation and I found a 'bug':
```
$ omero tag list
....
Traceback (most recent call last):
  File "/home/ldelisle/.conda/envs/omero/bin/omero", line 11, in <module>
    sys.exit(main())
  File "/home/ldelisle/.conda/envs/omero/lib/python3.6/site-packages/omero/main.py", line 125, in main
    rv = omero.cli.argv()
  File "/home/ldelisle/.conda/envs/omero/lib/python3.6/site-packages/omero/cli.py", line 1784, in argv
    cli.invoke(args[1:])
  File "/home/ldelisle/.conda/envs/omero/lib/python3.6/site-packages/omero/cli.py", line 1222, in invoke
    stop = self.onecmd(line, previous_args)
  File "/home/ldelisle/.conda/envs/omero/lib/python3.6/site-packages/omero/cli.py", line 1299, in onecmd
    self.execute(line, previous_args)
  File "/home/ldelisle/.conda/envs/omero/lib/python3.6/site-packages/omero/cli.py", line 1381, in execute
    args.func(args)
  File "/home/ldelisle/.conda/envs/omero/lib/python3.6/site-packages/omero/plugins/tag.py", line 715, in list
    self.pagetext(lines)
  File "/home/ldelisle/.conda/envs/omero/lib/python3.6/site-packages/omero/plugins/tag.py", line 206, in pagetext
    self.print_line(line, index)
  File "/home/ldelisle/.conda/envs/omero/lib/python3.6/site-packages/omero/plugins/tag.py", line 196, in print_line
    input = input("[Enter], [f]orward forever, or [q]uit: ")
UnboundLocalError: local variable 'input' referenced before assignment
```

PR #307, opened 2021-10-11
change `input` to `input_val`

lldelisle:
To solve #306


lldelisle:
Where should I send the form? Do you need a 'real' signature (print and scan) or just my name in the form is sufficient.

PR #309, opened 2021-11-29
Implementation for DatasetColumn

muhanadz:
Added implementation for `DatasetColumn` which was imported from `omero.grid` in `omero metadata` but was never implemented. `DatasetColumn` header defined [here](https://github.com/ome/omero-blitz/blob/bf2802814f4a1d5056c64ad825b583339f474ede/src/main/slice/omero/Tables.ice#L62). 

This PR hopes to start the process of supporting dataset IDs as column type (currently only dataset names are supported). 
`omero metadata` ready to handle `DatasetColumn` [here](https://github.com/ome/omero-metadata/blob/148b6a4d4e5cf1deb78e3f087b5907024c749828/src/omero_metadata/populate.py#L457)

Might need to add `DatasetColumn` test case [here](https://github.com/ome/openmicroscopy/blob/2a7ef41adda6bde343fae6db047c0737b8bde958/components/tools/OmeroPy/test/integration/tablestest/test_backwards_compatibility.py#L389) once merged.

### DIFF
--- a/src/omero/columns.py
+++ b/src/omero/columns.py
@@ -199,6 +199,15 @@ class PlateColumnI(AbstractColumn, omero.grid.PlateColumn):
     def descriptor(self, pos):
         return tables.Int64Col(pos=pos)
 
+class DatasetColumnI(AbstractColumn, omero.grid.DatasetColumn):
+
+    def __init__(self, name="Unknown", *args):
+        omero.grid.DatasetColumn.__init__(self, name, *args)
+        AbstractColumn.__init__(self)
+
+    def descriptor(self, pos):
+        return tables.Int64Col(pos=pos)
+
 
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks, @muhanadz. :+1: for getting a test in ome/openmicroscopy, both at that location but perhaps more generally with the use-case that you're covering. The test can still go green even before this PR is merged since all PRs are merged together daily for testing purposes:

 * https://github.com/snoopycrimecop/omero-py/tree/merge_ci
 * https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/

joshmoore:
Thanks for slogging it through, @muhanadz! https://github.com/ome/openmicroscopy/pull/6300 is green and in. :heart:

PR #310, opened 2021-12-06
Decode script stdout/stderr before printing

jmuhlich:
This change fixes a `bytes`-`str` concatenation issue in `omero script launch` output under Python 3 that caused the stdout/stderr lines to be printed as their `repr`. Output that previously looked like this:
```
	* b'Image:1 : Processing...\n--------------------\n'
```
Now looks correct:
```
	* Image:1 : Processing...
	* --------------------
	* 
```
(Tracebacks are now readable again too)

### DIFF
--- a/src/omero/plugins/script.py
+++ b/src/omero/plugins/script.py
@@ -436,7 +436,7 @@ class ScriptControl(BaseControl):
         def p(m):
             class handle(object):
                 def write(this, val):
-                    val = "\t* %s" % val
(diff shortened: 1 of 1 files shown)


will-moore:
Thanks for the fix. Working fine for me. 👍

sbesson:
Should we get this into a quick 5.10.2 release?

joshmoore:
Sounds good.

PR #311, opened 2021-12-13
fix typo

jburel:
Fixed typo found when investigating a parsing issue


joshmoore:
Oh wow. Not sure of what the impact of this might be! All for getting this in, but let's see how merge-ci and latest-ci differ tomorrow.

will-moore:
I don't see any impact as `len("omero.version=")` is the same length as `len("omero.verison=")`.

joshmoore:
Ah cheers. Just saw this go by during a call and didn't noticed the `len()`. Merging.

PR #312, opened 2021-12-13
Config output exclude versions

jburel:
Exclude keys with ``version`` from the output of the ``config parse`` command

cc @sbesson @joshmoore

### DIFF
--- a/src/omero/install/config_parser.py
+++ b/src/omero/install/config_parser.py
@@ -169,11 +169,9 @@ HEADER = \
 
 %(properties)s"""
 
-BLACK_LIST = ("##", "versions", "omero.upgrades")
+BLACK_LIST = ("##", "versions", "omero.upgrades", "omero.version")
 
 STOP = "### END"
-
-EXCLUDED_VERSION = ["omero.version"]
 
 import os
 import argparse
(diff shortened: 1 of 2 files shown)


src/omero/install/config_parser.py
```diff
@@ -173,6 +173,8 @@ def get_reference(self):
 
 STOP = "### END"
 
+VERSION = "version"
```
joshmoore, line 176: Are the other version properties showing up? If not, it might be best to be explicit about `omero.version` since this might be tricky to detect if we were to add `omero.bf_version` with an auto-download of the jar, etc.

jburel: I had that in place initially, we currently have ``omero.db.version`` in the list.

src/omero/install/config_parser.py
```diff
@@ -173,6 +173,8 @@ def get_reference(self):
 
 STOP = "### END"
 
+EXCLUDED_VERSION = ["omero.version"]
```
sbesson, line 176: What is the different between this new API and  `BLACK_LIST` defined four lines above. Would appending `omero.version` to the set suffice to exclude this property from being parsed?

(NB: as per our recent review, this might be the occasion to do `s/BLACK_LIST/EXCLUDE_LIST`

jburel: I did not want to confuse things but happy to revisit it

sbesson: If the two exclusion lists achieve different goals, it's fine to introduce new logic. But even then it might be useful to know and document in which context each list should be used when excluding new properties.

jburel: Looking at closely, they have similar goals so I have added to the list
I have renamed the list. I have **not** renamed the method but happy to do so

sbesson: Yes. Let's keep things contained and get this out as a patch release. I will open a follow-up PR to rename the method and deprecate the old one

sbesson (approved):

PR #319, opened 2022-02-22
update year

jburel:
Nothing to test


sbesson (approved):
Same considerations as https://github.com/ome/omero-web/pull/360#pullrequestreview-889687908

PR #321, opened 2022-03-09
Don't ignore params limit if offset is None

### LINKED ISSUE #320: Bug: getObjects() params ignores limit if offset is None
Found by @erickmartins at https://github.com/ome/omero-gallery/pull/94

`conn.getObjects()` allows each `ObjectWrapper` to provide a `params` object,
allowing filtering based on object properties.
We use various attributes of the params provided in `conn.getObject("Image", params)` and apply these to the params from
the ObjectWrapper (although we only handle pagination and ownerId).

However, the pagination (offset and limit) is only applied if BOTH are set:

https://github.com/ome/omero-py/blob/bf230dd4230215d91300f110f625a9a4adf3f5ee/src/omero/gateway/__init__.py#L3384

We should simply set offset and limit individually.

will-moore:
Fixes #320 

To test:

The params `limit` should not be ignored:
```
params = omero.sys.ParametersI()
params.theFilter = omero.sys.Filter()
params.theFilter.limit = wrap(1)
images = list(conn.getObjects("Image", params=params))
assert len(images) == 1
```

I'll add an integration test...

cc @erickmartins @joshmoore

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -3386,11 +3386,10 @@ class _BlitzGateway (object):
             if lmt is not None:
                 limit = lmt.val
 
-        baseParams.theFilter = omero.sys.Filter()
         if limit is not None:
-            baseParams.theFilter.limit = rint(limit)
-        if offset is not None:
-            baseParams.theFilter.offset = rint(offset)
+            if offset is None:
+                offset = 0
+            baseParams.page(offset, limit)
 
         # getting object by ids
         if ids is not None:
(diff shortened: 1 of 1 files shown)


joshmoore:
I don't assume anyone is depending on this behavior, so :+1: but in general, calling `page(0, limit)` would achieve the same, right?

joshmoore:
This looks good to me. :+1: There may be room for clarification in the doc string to specify what takes priority, etc. but that's not directly related to @erickmartins' issue.

will-moore:
On `getObjects()` and `buildQuery()` we have:
```
 :param params:      omero.sys.Parameters, can be used for pagination,
                            & filtering by owner. Takes precedence over opts.
```

jburel:
Thanks
Suggesting to merge and tag 5.11.1

PR #326, opened 2022-05-19
Windows import fix

sbesson:
See https://forum.image.sc/t/omero-cli-on-windows-for-image-import/67248 for a description of the issue

As suggested by @manics, `zipfile.Zipfile` uses `/` as the path separator agnosticly of the underlying platform.
1d99f7e71dcc37cb4ce16ff30d1ce9bb4c52ddbe updates the code accordingly and should fix the JAR fetching and hence the import on Windows systems.

Incidentally, this fix seems to be sufficient to re-enable all CLI import tests on Windows with an extra commit sanitizing the stderr output for the `assert` comparisons.

Proposed tag: `5.11.2`

PR #327, opened 2022-05-20
Method to add time to ParametersI

muhanadz:
Added new addTime() method for `omero.sys.ParametersI` class. Added tests to test this method.

Without this PR, adding `rtime` type object required manual mapping:
```python
params = omero.sys.ParametersI()
date = datetime.datetime.fromtimestamp(1)
params.map["Date"] = omero.rtypes.rtime(date)
```
With this PR, it is possible to add the date directly:
```python
params = omero.sys.ParametersI()
date = datetime.datetime.fromtimestamp(1)
params.addTime("Date", date)
```

### DIFF
--- a/src/omero_sys_ParametersI.py
+++ b/src/omero_sys_ParametersI.py
@@ -359,4 +359,8 @@ class ParametersI(omero.sys.Parameters):
         self.add(name, rstring(stringValue))
         return self
 
+    def addTime(self, name, TimeValue):
+        self.add(name, rtime(TimeValue))
+        return self
+
 _omero_sys.ParametersI = ParametersI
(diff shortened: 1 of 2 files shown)


sbesson (approved):
The API addition looks straightforward and mirrors the existing convenience methods for passing primary types. The change will require a minor release incremental to mark the backwards-compatible API and potentially could go in with #326.

@muhanadz is there a location in the OMERO.py/OMERO.web that could make use of this new API?

muhanadz:
@sbesson I'm not sure. I simply needed to add time to an HQL query via the ParametersI() model but instead had to manually map the datetime object, since there wasn't a method for that. Added this method in case someone needed this in the future.

PR #328, opened 2022-05-24
test_config: use native tmp_path pytest fixture

### LINKED ISSUE #42: Concurrency issue(s) with tox?
```
[gw0] Python 3.6.3 (default, May 31 2019, 13:05:43)  -- [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
[gw1] Python 3.6.3 (default, May 31 2019, 13:05:43)  -- [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
[gw2] Python 3.6.3 (default, May 31 2019, 13:05:43)  -- [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
[gw3] Python 3.6.3 (default, May 31 2019, 13:05:43)  -- [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
gw0 ok / gw1 ok / gw2 ok / gw3 ok
――――――――――――――――――――――――――――――――――――――――――――――― test/unit/tablestest/test_hdfstorage.py ―――――――――――――――――――――――――――――――――――――――――――――――
test/unit/tablestest/test_hdfstorage.py:28: in <module>
    from library import TestCase
test/unit/tablestest/library.py:14: in <module>
    from omero.util.temp_files import create_path
target/omero/util/temp_files.py:349: in <module>
    manager = TempFileManager()
target/omero/util/temp_files.py:65: in __init__
    if not self.create(self.userdir) and not self.access(self.userdir):
target/omero/util/temp_files.py:246: in create
    dir.makedirs(0o700)
target/path.py:1253: in makedirs
    os.makedirs(self, mode)
.tox/py36/lib64/python3.6/os.py:220: in makedirs
    mkdir(name, mode)
E   FileExistsError: [Errno 17] File exists: path('/home/tox/omero/tmp/omero_tox')
```

sbesson:
Fixes #42 (see https://github.com/ome/omero-py/runs/6574741655?check_suite_focus=true for the latest occurence of this bug)

The current `test_config` implementation makes use of the temporary file utility the `omero.util.temp_files` module. This implementation is not thread safe when running the pytest in a distributed manner as the username ("runner") used for the based directory is the same for all workers.
This creates a potential race condition when  `from omero.util.temp_files import create_path` is invoked concurrently as multiple workers might attempt to use `os.makedirs` to create the same `omero_runner` directory.

This PR updates the test implementation to use pytest native [tmp_path fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html). Each temporary directory should be unique within the scope of the test so there should be no conflict between workers.

### DIFF
--- a/test/unit/test_config.py
+++ b/test/unit/test_config.py
@@ -19,7 +19,6 @@ import json
 import errno
 import pytest
 from omero.config import ConfigXml, xml
-from omero.util.temp_files import create_path
 from omero_ext import portalocker
 
 from xml.etree.ElementTree import XML, Element, SubElement, tostring
@@ -100,19 +99,19 @@ def totext(elem):
 
 class TestConfig(object):
 
-    def testBasic(self):
-        p = create_path()
+    def testBasic(self, tmp_path):
+        p = tmp_path / "config.xml"
         config = ConfigXml(filename=str(p))
         config.close()
-        assertXml(initial(), XML(p.text()))
+        assertXml(initial(), XML(p.read_text()))
 
-    def testWithEnv(self):
-        p = create_path()
+    def testWithEnv(self, tmp_path):
+        p = tmp_path / "config.xml"
         config = ConfigXml(filename=str(p), env_config="FOO")
         config.close()
-        assertXml(initial("FOO"), XML(p.text()))
+        assertXml(initial("FOO"), XML(p.read_text()))
 
-    def testWithEnvThenWithoutEnv(self):
+    def testWithEnvThenWithoutEnv(self, tmp_path):
         """
         This test shows that if you create the config using an env
         setting, then without, the __ACTIVE__ block should reflect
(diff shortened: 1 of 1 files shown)


joshmoore (approved):
Interesting & good find. I do wonder about the possible issues that `create_path`'s non-thread-safety might cause.

This has been in for some time and things are green, :+1:

sbesson:
> I do wonder about the possible issues that create_path's non-thread-safety might cause.

Hopefully the impact should be limited as the issue is related to the  creation of the parent directory `<OMERO_TEMPIR>/omero_<user>`. Since this parent directory is not cleaned up by the library, I think we are dealing with a one-off race condition per user. If this was causing problems with the library usage (rather than the tests), a possible workaround would be to pass `exist_ok=True` to `os.makedirs` in `TempFileManager.__init__`

PR #329, opened 2022-05-25
Also pass omero.db.properties to reindex

chris-allan:
Under certain conditions, in particular those where a specific `sslmode` [^1] is required, `omero.db.properties` is an essential property that must be provided for the reindexing code to be able to communicate with the database.

[^1]: https://jdbc.postgresql.org/documentation/head/ssl-client.html

### DIFF
--- a/src/omero/plugins/admin.py
+++ b/src/omero/plugins/admin.py
@@ -1817,7 +1817,7 @@ present, the user will enter a console""")
         cfg = config.as_map()
         omero_data_dir = self._get_data_dir(config)
         config.close()  # Early close. See #9800
-        for x in ("name", "user", "host", "port"):
+        for x in ("name", "user", "host", "port", "properties"):
             # NOT passing password on command-line
             k = "omero.db.%s" % x
             if k in cfg:


chris-allan:
@stick will validate this on our side. We have a user who has this requirement in production (`sslmode=require`) and currently can't reindex because of it.

Couple other thoughts which are possibly beyond the scope of this PR:

1. If a user is making use of `omero.db.url` [^1] instead of the individual properties reindexing won't work
2. If `omero.db.pass` is not being provided on the command line I'm not sure how reindexing works if a password other than `omero` is being used (@stick and I will validate this)

[^1]: https://docs.openmicroscopy.org/omero/5.6.4/sysadmins/config.html#omero-db-url

stick (approved):
tested internally and all seems to work as desired.
various db.properties are honored as expected.

joshmoore (approved):
:+1:

sbesson (commented):
Also tested  `omero  admin reindex` with a default server setting and confirmed the behavior is unafffected. Overall, the additions are sensible and this should be ready to include in the upcoming OMERO.py release. Is there a timeline for having this fix released from your side?

chris-allan:
> Is there a timeline for having this fix released from your side?

None right now. We managed to work around the problem by forcing TLS 1.2 (which is incidentally what the user really wanted anyway).

will-moore:
FYI: this is now in the recent 5.12.0 release of `omero-py`

PR #331, opened 2022-08-18
setChannelNames performance

will-moore:
See https://forum.image.sc/t/apply-channel-names-to-all-images-in-an-omero-dataset-possible-bug/70521/12

It seems that `conn.setChannelNames()` is very slow in cross-group queries (which is the default behaviour for omero-web) and this is due to `containerService.getImages("Dataset", ids)`.
This PR should improve the speed by using a much faster alternative. See forum above for performance comparison.

Should also help fix other issues with setChannelNames: https://github.com/ome/omero-web/issues/304

To test:
 - Rename channels in a Dataset
 - Should work as before, but feel a bit faster

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -3964,9 +3964,10 @@ class _BlitzGateway (object):
         if data_type == "Image":
             imageIds = [int(i) for i in ids]
         elif data_type == "Dataset":
-            images = self.getContainerService().getImages(
-                "Dataset", ids, None, self.SERVICE_OPTS)
-            imageIds = [i.getId().getValue() for i in images]
+            imageIds = []
+            for dataset_id in ids:
+                images = self.getObjects("Image", opts={'dataset': dataset_id})
+                imageIds.extend([i.id for i in images])
         elif data_type == "Plate":
             imageIds = []
             plates = self.getObjects("Plate", ids)


will-moore:
Excluding to test performance without this PR on merge-ci

--exclude

pwalczysko:
With this PR

2000-images Datatset ID 43169

```
15987 ms
16067 ms
15845 ms
16032 ms
```

will-moore:
Using the code example from the forum above on that merge-ci Dataset, I see that new `getObjects()` is 1.7 seconds faster:
```
containerService.getImages() 0:00:02.902359
getObjects() 0:00:01.171373
```
In the case above, the querying of images (the step that is improved in this PR) is not the rate-limiting step, so only a small improvement may be observed.

pwalczysko:
In summary, the improvement observed in the UI and the script are roughly matching, the improvement is in the desired direction. 
Further, the functionality still works as expected. LGTM

pwalczysko (approved):

joshmoore:
This is currently excluded. Worth getting it back in the build?

will-moore:
Yes, I just removed the exclude label earlier...

src/omero/gateway/__init__.py
```diff
@@ -3964,9 +3964,10 @@ def setChannelNames(self, data_type, ids, nameDict, channelCount=None):
         if data_type == "Image":
             imageIds = [int(i) for i in ids]
         elif data_type == "Dataset":
-            images = self.getContainerService().getImages(
-                "Dataset", ids, None, self.SERVICE_OPTS)
-            imageIds = [i.getId().getValue() for i in images]
+            imageIds = []
+            for dataset_id in ids:
+                images = self.getObjects("Image", opts={'dataset': dataset_id})
+                imageIds.extend([i.id for i in images])
```
jburel, line 3970: https://github.com/ome/omero-web/pull/394#issuecomment-1237978346 Renaming of plates is directly using ``getObjects`` i.e. 
```
qs = self.getQueryService()
result = qs.findAllByQuery(query, params, self.SERVICE_OPTS)
```
The ``getContainerService().getImages`` loads more information about the image object itself i.e. it leads to more ``join``. This call only needs the image ID so switch to ``getObjects`  makes sense

pwalczysko:
We have a pre-deployment test of channel renaming on ns

1. find an svs dataset with 3 images logged in as full admin
2. try to rename the channels
3. observe that the action expires and no renaming happens

I will add here the status once this PR has beeen deployed on the ns production server.

pwalczysko:
Major success, see the report in https://github.com/ome/omero-web/pull/394#issuecomment-1243718199

PR #332, opened 2022-08-31
Build conda

jburel:
This PR introduces a job testing the installation of omero-py in a conda env and connect to IDR
The installation is done on MAc, Windows and Ubuntu. This follows a post on image.sc
Open against this repo but https://github.com/conda-forge/omero-py-feedstock could be considered

cc @dominikl


joshmoore:
* Big :+1: for testing conda.
* Pretty strong preference for the feedstock, since changes there wouldn't be tested here (but vice versa works)
* A bit uncertain about testing against IDR.

jburel:
We could use another server but I did not want to introduce a secret param

joshmoore:
Sorry for having dropped this. Should we discuss maybe having a read-only user for demo? The two positives I can think of:
 - It would keep us regularly testing demo (and alert us if it has fallen over).
 - The SLA specifies that if the tests cause it to fall over, well so be it.

jburel:
I forgot about it too.
Having a read-only user will work

connect.py
```diff
@@ -0,0 +1,11 @@
+import sys
+from omero.gateway import BlitzGateway
+HOST = 'ws://idr.openmicroscopy.org/omero-ws'
```
joshmoore, line 3: My concerns here persist.

jburel: not suggesting to merge as it is. I will discuss post migration to RHEL 9 the demo option with @pwalczysko 
I was just using it to validate.

joshmoore: :+1:

sbesson:
Should this be closed in favor of a PR against https://github.com/conda-forge/omero-py-feedstock as this is fairly specific to the Conda distribution mechanism?

joshmoore:
I'd guess the conda-forge org might be more limited in terms of GHA computational, but at least there are far fewer PRs there.

manics:
You can run test scripts in the conda recipe, it shouldn't need a separate workflow:
https://github.com/conda-forge/omero-py-feedstock/blob/20d9f0ca55b3f7472cfe5a5171cc33e969bf0695/recipe/meta.yaml#L56-L58
https://conda-forge.org/docs/maintainer/adding_pkgs/#copying-test-files

PR #335, opened 2022-10-03
Always report pixeldata time in the output of fs importtime

sbesson:
While investigating, mismatches between the sum of the import times reported by the CLI command and the import logs, I realised the time for completing this step is being skipped if no rendering settings are assigned. In addition to pyramid generation, several actions are being performed including min/max calculation and pixel/series assignment.

## Testing

Without this PR, import a file e.g. a fake image/plate with and without setting the thumbnails

```
omero import "test&plateRows=4&plateCols=4.fake"
omero import "test&plateRows=4&plateCols=4.fake" --skip thumbnails
```

Then run `omero fs importtime Fileset:<id>` on both filesets. Without this PR the `pixeldata` step of the second command should be omitted from the output. With this PR included, it should be reported in both cases

### DIFF
--- a/src/omero/plugins/fs.py
+++ b/src/omero/plugins/fs.py
@@ -1263,6 +1263,7 @@ class ImportTime(object):
         self.metrics['UPLOAD'] = upload_end - upload_start
         self.metrics['SET_ID'] = set_id_end - upload_end
         self.metrics['METADATA'] = metadata_end - set_id_end
+        self.metrics['PIXELDATA'] = pixeldata_end - metadata_end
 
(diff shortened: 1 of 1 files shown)


joshmoore (approved):
MSTM

sbesson:
@jburel @joshmoore thoughts about scheduling an immediate `5.12.1` release including this bug fix or do you have other changes you would like to see included?

joshmoore:
None from me.

jburel:
No other changes from my side after quickly reviewing the open PRs

PR #337, opened 2022-10-06
bulk.yml depth is passed to xargs

will-moore:
When performing bulk import, any `depth: 100` value added to the `bulk.yml` is currently ignored and you have to use `--depth=100` on the command line.

This attempts to fix that issue.

This fix works, BUT it now overrides any value passed on the command-line (which probably should take precedence)?
However, at this point in the code, I can't check `xarg` for the presence of any `-Domero.import.depth` command, since this is always there with the default of 4.

Not quite sure where the "right" place is to fix this?

cc @sbesson @joshmoore @dominikl

### DIFF
--- a/src/omero/plugins/import.py
+++ b/src/omero/plugins/import.py
@@ -672,6 +672,8 @@ class ImportControl(BaseControl):
                         mode = "w"
                     else:
                         mode = "a"
+                    if command_args.depth:
+                        xargs.append("-Domero.import.depth=%s" % command_args.depth)
                     self.do_import(command_args, xargs, mode=mode)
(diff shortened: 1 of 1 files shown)


sbesson:
Looking briefly at the way the API are called, either the full `xargs` set-up or maybe only the `omero.import.depth` set-up could potentially be moved under `do_import` so that it's only defined once?

I would also have assumed that arguments passed via the command-line would take precedence as well but looking at the implementation, it is possible that this might not be the case as the parsing of the bulk keys calls `CommandArguments.add` for most keys and I do not see the logic for handling existing keys. Additional unit tests might be needed to confirm this.

PR #348, opened 2022-11-17
add env variable to fix doc generation

jburel:
Workaround discussed so we can generate the API doc.
Env variable set in readthedocs

### DIFF
--- a/src/omero/util/temp_files.py
+++ b/src/omero/util/temp_files.py
@@ -346,7 +346,7 @@ class TempFileManager(object):
         self.logger.error(
             "rmtree error: %s('%s') => %s", func.__name__, name, exc[1])
 
-if "NO_TEMP_MANAGER" not in os.environ:
+if "OMERO_NO_TEMP_MANAGER" not in os.environ:
     manager = TempFileManager()
 
 """
--- a/src/omero/util/temp_files.py
+++ b/src/omero/util/temp_files.py
@@ -346,8 +346,7 @@ class TempFileManager(object):
         self.logger.error(
             "rmtree error: %s('%s') => %s", func.__name__, name, exc[1])
 
-no_tmp_manager = os.environ.get('NO_TEMP_MANAGER', None)
-if no_tmp_manager is None:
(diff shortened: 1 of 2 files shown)


src/omero/util/temp_files.py
```diff
@@ -346,7 +346,10 @@ def on_rmtree(self, func, name, exc):
         self.logger.error(
             "rmtree error: %s('%s') => %s", func.__name__, name, exc[1])
 
-manager = TempFileManager()
+no_tmp_manager = os.environ.get('NO_TEMP_MANAGER', None)
```
joshmoore, line 349: I wouldn't add this to the namespace, J-m, unless you need to use it somewhere else. Otherwise it could (inadvertently) become part of the public API.

joshmoore, line 349: I wouldn't add this to the namespace, J-m, unless you need to use it somewhere else. Otherwise it could (inadvertently) become part of the public API.

joshmoore (commented):
Barring the one minor comment, this matches what we spoke about regarding the weirdness of the build. I'd still be very interested to know why `path()` is not behaving as we expect, but also big :+1: for getting the released docs updated.

jburel:
> Barring the one minor comment, this matches what we spoke about regarding the weirdness of the build. I'd still be very interested to know why `path()` is not behaving as we expect, but also big 👍 for getting the released docs updated.

I will look into it afterwards.

jburel:
https://omero-py--348.org.readthedocs.build/en/348/omero.util.temp_files.html with [78638a0](https://github.com/ome/omero-py/pull/348/commits/78638a04863cd78ca24fef30ce3f5aea93ca8cda)

jburel:
@joshmoore any more comments?

joshmoore:
:+1: The only thing I got caught up on was whether we should prefix with `OMERO_` etc. etc. But perhaps if it's a temporary solution (famous last words...), then it doesn't matter too much.

jburel:
I will add ``OMERO`` to avoid confusion

jburel:
@joshmoore done

PR #349, opened 2022-11-17
remove sphinx job after rtd migration

jburel:
The job should no longer be run


jburel:
Job is not run anymore. Merging

PR #336, opened 2022-10-04
Remove Anonymous Diffie-Hellman default configuration

chris-allan:
Since OMERO 5.6.2 (released in July, 2020) we have been recommending the use of `omero certificates` to ensure that all OMERO server installations have, at minimum, a self-signed certificate. Consequently, I think it's time we removed the Anonymous Diffie-Hellman (ADH) default configuration from omero-py.

We know that the detection is error prone and that ADH support is not present in newer OpenSSL versions.  So this change also helps us seamlessly support newer Linux distributions such as Ubuntu 22.04 as well as CentOS/RHEL/Rocky 9 and allows us to utilize manylinux wheels where OpenSSL version mismatches were making detection an ugly mess.

Cipher suites and minimum TLS version can still be enforced by server configuration.

PR #351, opened 2022-11-21
changelog for 5.13.0

jburel:
cc @sbesson @will-moore

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## New features
 
-- Remove Anonymous Diffie-Hellman default configuration [#336](https://github.com/ome/omero-py/pull/336)
+- Remove Anonymous Diffie-Hellman default configuration [#336](https://github.com/ome/omero-py/pull/336). This change will require to use of ``omero certificates`` to ensure that an OMERO server installation has, at minimum, a self-signed certificate.
 - Publish API doc [#330](https://github.com/ome/omero-py/pull/330)
 
 ## Bug fix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - omero admin diagnostics: move OMERO.py version as part of the output [#339](https://github.com/ome/omero-py/pull/339)
 - Close correct locker [#346](https://github.com/ome/omero-py/pull/346)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+# 5.13.0 (November 2022)
+
+## New features
+
+- Remove Anonymous Diffie-Hellman default configuration [#336](https://github.com/ome/omero-py/pull/336)
+- Publish API doc [#330](https://github.com/ome/omero-py/pull/330)
```
sbesson, line 6: Minor suggestion: https://github.com/ome/omero-py/blob/master/setup.py#L205 could be updated to also include the new API documentation location?

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+# 5.13.0 (November 2022)
+
+## New features
+
+- Remove Anonymous Diffie-Hellman default configuration [#336](https://github.com/ome/omero-py/pull/336)
```
sbesson, line 5: Is there a place that explains that the certificates MUST be configured server-side when upgrading toh this version of the OMERO.py client? Should this belong to the release notes

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+- Publish API doc [#330](https://github.com/ome/omero-py/pull/330)
+- Switch to new output command (GHA) [#340](https://github.com/ome/omero-py/pull/340)
```
sbesson, line 7: This does not feel like a feature but it's not really a bug fix either. We have used `Other updates` in the past, should we use this for general maintenance category?

sbesson (commented):
A few comments. Probably the biggest question is around being are explicit about the server steps required to work with the changes in #336

CHANGELOG.md
jburel: I will put it under ``other updates``

CHANGELOG.md
jburel: I have added info into changelog. But I think I will also add line in README

CHANGELOG.md
jburel: done

sbesson (approved):

PR #353, opened 2022-11-22
omero.clients: initialize connection retry reason out of the while loop

### LINKED ISSUE #352: Informative error reason set to None before sending to logger in clients.py
Thanks for supporting omero.

https://github.com/ome/omero-py/blob/master/src/omero/clients.py#L631

In `client.createSession()` if one of two Exceptions occurs, an informative `reason` for the exception is captured.  However, the value of `reason` is set to None before it is logged at https://github.com/ome/omero-py/blob/master/src/omero/clients.py#L633.  As a result, in log entries, the `reason` is always `None`, losing information about the reason for failure of `createSession()`.

In a recent debugging session of a connection timeout, being able to see the `reason` in the log would have been very helpful.

Suggestion: move assignment of `reason` at https://github.com/ome/omero-py/blob/master/src/omero/clients.py#L631 up a line so that it is outside the `while()` loop.

sbesson:
Fixes #352 

The logic should be unchanged by the exception shouldn't be reset to `None` and displayed on each retry

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -627,8 +627,8 @@ class BaseClient(object):
             # Acquire router and get the proxy
             prx = None
             retries = 0
+            reason = None
             while retries < 3:
-                reason = None
                 if retries > 0:
                     self.__logger.warning(
                         "%s - createSession retry: %s" % (reason, retries))


joshmoore (approved):

jburel:
Including this PR in the upcoming 5.13.
I will adjust the changelog accordingly

PR #354, opened 2022-11-22
Add inherited-members option to the gateway automodule configuration

sbesson:
As most of these classes are effectively aliases/inherited, the current generated API documentation is pretty bare for the omero.gateway package as most methods are inherited from private base classes. This commit investigates using the inherited-members option to expose the docstrings

### DIFF
--- a/docs/omero.gateway.rst
+++ b/docs/omero.gateway.rst
@@ -24,5 +24,6 @@ Module contents
 .. automodule:: gateway
    :members:
    :inherited-members:
+   :private-members:
    :undoc-members:
    :show-inheritance:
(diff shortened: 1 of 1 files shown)


sbesson:
@will-moore @jburel see https://omero-py--354.org.readthedocs.build/en/354/omero.gateway.html#gateway._BlitzGateway as a starting point of feedback and discussion.

The options add a fair amount of verbosity but something of that nature is necessary in order to expose the docstrings for pieces of functionality as central e.g. the BlitzGateway constructor. I believe this is primarily a consequence of the current design of the gateway where for instance the public `BlitzGateway` class inherits from a private `_BlitzGateway` class where all the public API methods are defined

will-moore:
LGTM 👍

jburel (approved):
Patch release to have it under the stable docs asap?

jburel:
https://omero-py.readthedocs.io/en/stable/omero.gateway.html

Issue #360, opened 2023-01-17
zeroc-ice fails to build wheel


vpilling:
Can you please update the **omero-py** package to use the latest **zeroc-ice** package. The one that is currently included fails to compile and create wheel. The current version corrects the problem.

jburel:
By the latest do you mean 3.7.x?
omero does not support ice 3.7
I have bumped the version to use the latest version in the 3.6 series

michaelosthege:
Also that newer version is creating problems - I guess largely because https://github.com/conda-forge/zeroc-ice-feedstock isn't getting much attention and doesn't test builds on Windows.

Already with a simple `mamba create -n test -c conda-forge "python=3.10.x" omero-py`:
* With `python=3.10.8` it still works
* Any newer Python version results in either the dependency resolution not working, **or a broken `pip`**:

```
pip --version
>>> ImportError: DLL load failed while importing _ctypes: Das angegebene Modul wurde nicht gefunden.
```

markovendelin:
The problem is with zeroc-ice 3.6.5 failing to install when used with Python 3.11:

```
      src/Slice.cpp:24:10: fatal error: eval.h: No such file or directory
         24 | #include <eval.h>
            |          ^~~~~~~~
      compilation terminated.
```

It has been fixed in https://github.com/zeroc-ice/ice/pull/1394 (closing https://github.com/zeroc-ice/ice/issues/1393), but that seems not to be backported to 3.6.x. So, as it is, we cannot use omero-py on Python 3.11

jburel:
Alternatively you can try to install from the ome channel
``conda install -c ome zeroc-ice36-python==3.6.5``

markovendelin:
Any chance it would make it to PyPi? We are using omero-py as a dependency for one of our projects to support OMERO as an image server backend. However, we don't publish it on Conda and don't plan to use Conda either.

jburel:
@markovendelin A new wheel will need to be build, are you running on CentOS, Ubuntu?

markovendelin:
Ping. https://github.com/conda-forge/zeroc-ice-feedstock/tree/version_3.6 cannot be used with `pip` as 

```
> pip install git+https://github.com/conda-forge/zeroc-ice-feedstock@version_3.6
Collecting git+https://github.com/conda-forge/zeroc-ice-feedstock@version_3.6
[...]

PR #361, opened 2023-01-18
use ice 3.6.5

jburel:
Bump ice version to 3.6.5 see https://github.com/ome/omero-py/issues/360


jburel:
3.6.5 matches what we have elsewhere but not the bulk of the problem
I have also rolled back a change in the README when using conda since now conda-forge installed 3.7.x and not 3.6.x as previously

PR #362, opened 2023-01-23
Deprecate omero.install.python_warning module

sbesson:
Noticed while fixing a bug on the OMERO.web CLI plugin (see https://github.com/ome/omero-web/pull/436), this module is no longer relevant since OMERO.py 5.6.0 and the migration to Python 3.x


jburel (approved):

will-moore (approved):
LGTM 👍

PR #368, opened 2023-04-06
drop Python 3.7

jburel:
Drop support for Python 3.7

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -22,7 +22,7 @@ jobs:
         include:
           - python-version: '3.8'
             os: macos-latest
-          - python-version: '3.8'
+          - python-version: '3.7'
             os: windows-latest
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


sbesson (approved):

jburel:
``hdf5`` tests currently fail on windows when using Python 3.8+

jburel:
@sbesson [9cbe240](https://github.com/ome/omero-py/pull/368/commits/9cbe2402a288b86539071342e22c9be9000a61dc) uses the wheel produced by GS
The tests are still failing on Windows so I have not bumped to 3.8+
You approved it before my last commit (and the build was still red). Are you okay with the current state of the PR?

I will investigate the hdf issue in a follow up.
Maybe an opportunity to remove our old copy of ``portalocker``. The files stay locked leading to the failure with Python 3.8

sbesson (approved):
@jburel no objections to moving forward with this PR.
On the Python 3.7/Windows front, we definitely need to look into these issues separately. Removing our vendored copy of `portalocker` had been already mentioned in https://github.com/ome/omero-web/pull/415#issuecomment-1273335369 during the migration to upstream `concurrent-log-handler` in OMERO.web.

Using upstream portalocker and deprecating `omero_ext/portalocker` would be valuable. From a quick code review, the main locations which will need functional review are `omero config`, `omero admin`, temporary files as well as OMERO.tables.

PR #305, opened 2021-09-29
Ensure we are using pyopenssl (See ome/omero-py#240)

chris-allan:
As of (psf/requests#5443) released in requests 2.24.0 [1] the urllib3
monkeypatching in of pyopenssl now only happens if the SNI is
unavailable (mostly just Python 2).  The prevailing intent within
the requests project seems to be to avoid pyopenssl use where
possible.  Obviously, this is not what we need so we're going to
have to follow the requests and urllib3 advice to do it explictly.

I have chosen to only perform the explicit injection for the upgrade
check.  Other uses of requests which want to avoid the sort of weird
OpenSSL errors like those outlined in (ome/omero-py#240) will need to do
the same thing.

  1. https://github.com/psf/requests/blob/main/HISTORY.md#2240-2020-06-17

PR #371, opened 2023-06-03
cap urllib3 to avoid openssl issue

jburel:
Problem noticed when running the omero-server job on CentOS 7 with SCL Python 3.8
cc @sbesson @khaledk2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -220,6 +220,7 @@ setup(
     },
     python_requires='>=3',
     install_requires=[
+        'urllib3<2',
         'appdirs',
         'future',
         'numpy',


sbesson:
Cross-linking to https://github.com/ome/scc/pull/284 which patched a similar issue (in a build-only dependency) and https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html which explains the issues associated with `urllib3 2.x`.

This problem will primarily impact CentOS 7 deployments with SCL Python 3.8 and OpenSSL 1.0.x. Unfortunately I am not aware of a classifier that would detect this environment and fallback to `urllib3 1.x`. 

The primary downside of this change is that the capping `urrlib3` will happen across the board including in more recent distributions where OpenSSL 1.1.1 or even 3.x are shipped with Python 3.7+. Looking at other changes, it looks like `urllib3 2.x` also deprecated the `urllib3.contrib.pyopenssl` module which was used as the basis of https://github.com/ome/omero-py/pull/305.

In summary:

- this is yet another incentive to decide and communicate on the CentO7 support timeline - see https://github.com/ome/omero-certificates/pull/33 for another SSL issue related to this environment
- more investigation is required on the urllib3 2.x breaking changes and their impact on the OMERO Python stack

PR #376, opened 2023-07-05
Use Image.LANCZOS alias rather than Image.ANTIALIAS

sbesson:
These aliases are strictly equivalent but the latter has been deprecated in Pillow 9.1 and dropped in Pillow 10 - see https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html


joshmoore (approved):
:+1:

jburel (approved):
straightforward change

PR #377, opened 2023-07-10
Let Ice choose the default SSL protocols that are available

chris-allan:
This setting made sense when we were still supporting SSLv3 but that has been disabled by default since Ice 3.6:

 * https://doc.zeroc.com/ice/3.6/property-reference/icessl#id-.IceSSL.*v3.6-IceSSL.Protocols

Consequently, it's better to leave this setting as the default and let the server dictate what it expects since we do not have a complete overlap in TLS version support on the platforms we support (CentOS 7 does not support TLS 1.3).  This setting currently forbids clients from negotiating TLS 1.3 where it is available.  Ice will use sane defaults on these platforms.

### DIFF
--- a/src/omero/clients.py
+++ b/src/omero/clients.py
@@ -231,7 +231,6 @@ class BaseClient(object):
 
         self._optSetProp(id, "IceSSL.VerifyDepthMax", "6")
         self._optSetProp(id, "IceSSL.VerifyPeer", "0")
-        self._optSetProp(id, "IceSSL.Protocols", "tls1_0,tls1_1,tls1_2")
 
         # Setting block size
         self._optSetProp(


chris-allan:
Follow on from the cleanup started in #336. Should allow Python clients to connect to servers that require TLS 1.3 and not affect any existing clients that want to use TLS 1.2 or below.

sbesson (commented):
Tested against a set of Dockerized container environment under the following Linux distributions (CentOS 7, Rocky Linux 8, Rocky LInux 9, Ubuntu 20.04, Ubuntu 22.04).

On all Linux distributions except CentOS 7, the `omero certificates` command was supplemented with the following configuration calls to add TLS 1.3 to the list of SSL protocols

```
omero config set omero.glacier2.IceSSL.Protocols "TLS1_0,TLS1_1,TLS1_2,TLS1_3"
omero config set omero.glacier2.IceSSL.ProtocolVersionMax TLS1_3
```

Each environment was used as a client to test cross-distribution compatibility. A minimal connection test was established using

```python
c = omero.client(host=server, port=4064, args=[ '--IceSSL.Trace.Security=1']
c.createSession(username, password)
c.closeSession()
```

with the `IceSSL.Trace.Security` option enabling the selected protocol to be logged when creating the session. The following protocols were selected for all client/server combinations:

| Client / Server | CentOS 7 | Rocky 8 | Rocky 9 | Ubuntu 20.04 | Ubuntu 22.04 |
|-----------------|----------|---------|---------|--------------|--------------|
| CentOS 7        | TLS 1.2  | TLS 1.2 | TLS 1.2 | TLS 1.2      | TLS 1.2      |
| Rocky 8         | TLS 1.2  | TLS 1.3 | TLS 1.3 | TLS 1.3      | TLS 1.3      |
| Rocky 9         | TLS 1.2  | TLS 1.3 | TLS 1.3 | TLS 1.3      | TLS 1.3      |
| Ubuntu 20.04    | fail     | TLS 1.3 | TLS 1.3 | TLS 1.3      | TLS 1.3      |
| Ubuntu 22.04    | fail     | TLS 1.3 | TLS 1.3 | TLS 1.3      | TLS 1.3      |

The two failures are related to the issue discussed in https://forum.image.sc/t/omero-login-ssl-error-dh-key/79574.

PR #380, opened 2023-07-30
add getROIs

barrettMCW:
hey y'all! I was going to put a patch into our groups dev library but i figured I'd see if you wanted it. Since getROICount already gets the ROIs i just put the get part in a different method and made count just check the length of results.

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -10354,21 +10354,22 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
                 if ns == a['ns']:
                     return a['value']
 
-    def getROICount(self, shapeType=None, filterByCurrentUser=False):
+    def getROIs(self, shapeType=None, filterByCurrentUser=False):
         """
-        Count number of ROIs associated to an image
+        Gets ROIs associated to an image
 
         :param shapeType: Filter by shape type ("Rectangle",...).
         :param filterByCurrentUser: Whether or not to filter the count by
                                     the currently logged in user.
-        :return: Number of ROIs found for the currently logged in user if
-                 filterByCurrentUser is True, otherwise the total number
+        :return: list of ROIs found for the currently logged in user if
+                 filterByCurrentUser is True, otherwise the total rois
                  found.
         """
 
         # Create ROI shape validator (return True if at least one shape is
         # found)
         def isValidType(shape):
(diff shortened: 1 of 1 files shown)


will-moore:
Hi, and thanks for the suggestion.

Methods of the BlitzGateway (e.g. `getROIs()`) would normally be expected to return `BlitzObjectWrapper` objects, not e.g. `omero.model.Roi` but rather wrap the roi with `omero.gateway.RoiWrapper(conn, roi)`.

However, you wouldn't want to init all those objects when just counting ROIs as there's potential for `BlitzObjectWrapper` to load more data on init.
So you might want a `def _get_rois()` to return `omero.model.roi` objects as you currently have for `getROIs()`, use this for `getROICount()` and then wrap those ROIs to return Wrapper objects for `def getROIs()`.
Does that make sense?

will-moore:
Just added the --include label, since I noticed this wasn't being merged in the daily build: see https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/271/console

Then we can confirm tests are still passing at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.test_rois/

In due course, it would also be nice to add tests for `conn.getROIs()` to https://github.com/ome/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/test_rois.py

src/omero/gateway/__init__.py
```diff
@@ -10385,6 +10374,40 @@ def isValidROI(roi):
...
+        Gets ROIs associated to an image
+
+        :param shapeType: Filter by shape type ("Rectangle",...).
+        :param filterByCurrentUser: Whether or not to filter the count by
+                                    the currently logged in user.
+        :return: list of ROIs found for the currently logged in user if
+                 filterByCurrentUser is True, otherwise the total rois
+                 found.
+        """
+        return [RoiWrapper(self.conn, roi) for roi in
```
will-moore, line 10396: This needs to be `self._conn` (missing underscore).

barrettMCW: Sorry about that and the delay to fix, been OOO. should be good now. I wouldn't mind writing up a little test in  the openmicroscopy repo if you guys want me to. thanks y'all!

joshmoore (approved):

PR #382, opened 2023-08-25
omero admin start: add warning for deprecated TLS protocols

sbesson:
The message should be raised either if omero.glacier2.IceSSL.Protocols is unset or if TLS 1.0 or 1.1 are found in the list of allowed protocols

Implements the suggestion in https://github.com/ome/omero-certificates/pull/36#issuecomment-1691376331.
The warning refers to the server upgrade page for the instructions but happy to use another reference page if needed.


jburel (approved):
Merging and tagging 5.16.1

PR #383, opened 2023-08-28
add missing config file

jburel:
Add requirement file for rtd


jburel:
Green. Merging so we can fix other PRs

PR #385, opened 2023-10-25
Add support for Python 3.11

sbesson:
Python 3.11 has been released over a year ago and is available across several operating systems. This PR includes a minimal set of changes allowing OMERO.py to be functinal with Python 3.11. 

## Summary of changes

The biggest impact of the Python 3.11 was found to be in the dependencies of OMERO.py rather than the codebase itself

- e1d0fa409ce23597ad0338850d09368b49350baa adds the Python 3.11 manylinux wheel package to the Tox framework and adds the `py311` environment to `tox.ini` and the GitHub workflow definition to allow the testing under Python 3.11
- the vendored fork of `path.py` uses the "universal newline" opening mode `open(<file>, 'U')` deprecated in Python 3.3. and removed in Python 3.11 [^1] . Eventually, there is probably a larger concerted effort to remove the usage of this fork in favor of `pathlib` available from the standard library. Within the context of this work, this was assessed to be a larger body of work. Instead e1d0fa409ce23597ad0338850d09368b49350baa simply adjusts `omero_ext/path.py` to use `r` since "universal newline" mode is used by default since Python 3.
- the `mox3` dependency [^2] used by the unit tests was an unofficial port of the Google mox framework to Python 3. This library is unmaintained, has seen no releases since 3+ years and its usage was found to be incompatible with some of the Python 3.11 changes. 614a2e2cb0904fe9514353281d9bd1a0bd23a8a5 deprecates its usage in `testlib` and 986aa8421d09b6fbfb7339837b27f0b23ac60cec adjusts several tests to use `pytest-mock`  [^3], a thin-wrapper of the `mock` library available under the standard Python library

## Testing

With this PR included, all CI builds and tests should keep passing. In addition, tt should be possible to use OMERO.py as a client as well as deploying OMERO.server in a Python 3.11 virtual environment.

## Impact & open questions

- 614a2e2cb0904fe9514353281d9bd1a0bd23a8a5 directly impacts OMERO.py integration tests maintained under https://github.com/ome/openmicroscopy. If no objection, I can open a companion Pull Request with the corresponding adjustments to use `pytest-mock`. Unlike the GitHub worklow builds, this might require additional infrastructure work to install the new dependency in the Jenkins CI environment where the integration tests are executed
[...]

PR #386, opened 2023-10-30
Unit tests: clean up warnings

sbesson:
Noticed while working on #385, the current unit tests throw a few warnings post-completion.
This should update the corresponding unit tests accordingly:

- for session tests where existing clients are expected not to be re-usable (due to conflicting arguments), the existing client is cleared from the store to comply with the deletion semantics
- table tests for column names throwing a `tables.NonNaturalNameWarnings` are now internally recording the warnings and testing them

CI tests should be unaffected but warnings should no longer be thrown after the tests report.


jburel (approved):
Thanks
Everything is green

PR #387, opened 2023-12-13
Numpy pixels type

will-moore:
Currently, if we want to find the numpy pixels type for an Image, we have to duplicate code from the BlitzGateway, as in https://github.com/ome/omero-cli-zarr/pull/157/files

This PR aims to avoid that duplication by providing `PixelsWrapper.get_numpy_pixels_type()`


src/omero/gateway/__init__.py
```diff
@@ -7537,7 +7541,7 @@ def getTiles(self, zctTileList):
                     planeX = width
                 # +str(sizeX*sizeY)+pythonTypes[pixelType]
                 convertType = '>%d%s' % (
-                    (planeY*planeX), pixelTypes[pixelType][0])
+                    (planeY*planeX), pixelTypes[pixelType])
```
sbesson, line 7544: Could we simplify this further by using `numpy.dtype(numpyType).char`?

src/omero/gateway/__init__.py
```diff
@@ -7495,6 +7496,10 @@ def getPlane(self, theZ=0, theC=0, theT=0):
         planeList = list(self.getPlanes([(theZ, theC, theT)]))
         return planeList[0]
 
+    def get_numpy_pixels_type(self):
```
sbesson, line 7499: This could use a docstring. Additionally, should this be simplified as `get_numpy_type()` ?

src/omero/gateway/__init__.py
```diff
@@ -55,6 +56,16 @@
 
 logger = logging.getLogger(__name__)
 THISPATH = os.path.dirname(os.path.abspath(__file__))
+OMERO_NUMPY_DTYPES = {
```
sbesson, line 59: I would call this `OMERO_NUMPY_TYPES` to avoid the `type` vs `dtype` confusion

sbesson (changes requested):

sbesson:
@will-moore would you consider this is still in scope for the next minor release of OMERO.py? If so, do you have a timeline for addressing the review comments?

src/omero/gateway/__init__.py
will-moore: I went for `get_numpy_dtype()` here since `dtype` is what they're called for numpy.

will-moore:
Thanks for the ping @sbesson - pushed some fixes.

src/omero/gateway/__init__.py
```diff
@@ -7495,6 +7496,16 @@ def getPlane(self, theZ=0, theC=0, theT=0):
...
 
+    def get_numpy_dtype(self):
+        """
+        Returns the numpy dtype corresponding to the OMERO pixels type.
+
+        This is the numpy dtype that is used for getTiles() and getPlane()
+        :return:    numpy dtype
+        """
+        pixels_type = self.getPixelsType().value
+        return OMERO_NUMPY_TYPES[pixels_type]
```
sbesson, line 7507: Should this wrap `return dtype(OMERO_NUMPY_TYPES[pixels_type])` as per the new method name?

src/omero/gateway/__init__.py
```diff
@@ -7504,22 +7515,13 @@ def getTiles(self, zctTileList):
...
-                      PixelsTypeint32: ['i', numpy.int32],
-                      PixelsTypeuint32: ['I', numpy.uint32],
-                      PixelsTypefloat: ['f', numpy.float32],
-                      PixelsTypedouble: ['d', numpy.float64]}
         rawPixelsStore = None
         sizeX = self.sizeX
         sizeY = self.sizeY
         pixelType = self.getPixelsType().value
-        numpyType = pixelTypes[pixelType][1]
+        numpyType = self.get_numpy_pixels_type()
```
sbesson, line 7524: The method name has been changed in the last commit

src/omero/gateway/__init__.py
will-moore: No, I'll rename again! Kinda confusing how `np.int8` isn't a dtype but a class. 4600d63

sbesson (approved):
Retested the last few commits using the OMERO.py gateway as a client against various images of different dimensionalities. In all cases, `PixelsWrapper.get_numpy_type()` returned the expected scalar type and consumer methods like `getTiles` kept working as expected.

Waiting for a round of successful CI builds including this PR before merging this. As discussed in the weekly OME meeting, the proposal is to create a new minor release of OMERO.py focused on Numpy improvements immediately after

PR #388, opened 2024-01-10
Fix issue with font

jburel:
Problem noticed while working on https://github.com/ome/omero-scripts/pull/214

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -31,7 +31,7 @@ jobs:
           py=$(echo ${{ matrix.python-version }} | tr -d .)
           echo "py=$py" >> $GITHUB_OUTPUT
         shell: bash
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
@@ -49,7 +49,7 @@ jobs:
       - test
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: actions/setup-python@v4
       - name: Build package
         run: |
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -51,6 +51,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
       - name: Build package
         run: |
           python -mpip install wheel
--- a/setup.py
+++ b/setup.py
@@ -224,7 +224,7 @@ setup(
         'appdirs',
         'future',
         'numpy',
-        'Pillow',
+        'Pillow>=10.0.0',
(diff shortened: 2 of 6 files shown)


jburel:
https://github.com/ome/omero-scripts/pull/214 is now green with that set of change via https://github.com/jburel/omero-test-infra/tree/test_omero-py

sbesson (commented):
Looks related to https://github.com/ome/omero-py/pull/376 as these APIs were removed in Pillow 10 - https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html#font-size-and-offset-methods

Thinking about it, is it also a good occasion to update `setup.py` to ensure a minimal Pillow version? These functions have been deprecated since Pillow 9.2.0 - https://pillow.readthedocs.io/en/stable/releasenotes/9.2.0.html#font-size-and-offset-methods

jburel:
since Pillow 10 has been around for over 6 months (01/07/2023). I think we should have to 10.0.0 as a minimum

sbesson:
Thanks @jburel . Requiring `Pillow>10` effectively enforces Python 3.8 as the minimal supported version in upcoming OMERO.py versions - see https://pillow.readthedocs.io/en/latest/installation.html#python-support

Given the current timeline for Python 3.x support and the fact Python 3.7 and earlier is EOL, no objections with enforcing this from my side

jburel:
@will-moore any comment from your side? It will be good to have that PR merged and do a release so we can move forward with the test-infra upgrade work

will-moore (approved):
Looks good.
When looking up `font.getbbox()` I noticed that it returns `x, y, x2, y2` not `x, y, width, height`. Width is the same as `x2` if the `x, y` origin is `0, 0`, but this isn't always true.
In this example, x is 0 but y isn't:

https://stackoverflow.com/questions/74327497/understanding-the-getbbox-method-of-a-pillow-font-object

However, that's only the case if stroke isn't zero, and even then x is still 0 so we're OK for width.
I tried to run `image.createMovie()` to test, but couldn't get that to work due to unrelated issues.
But I think the changes look good 👍

jburel:
@will-moore do you suggest to do ``box = font.getbbox(...) `` then ``box[2]-box[0]`` for the width to be on the safe side.?
This is no the case in the changes you made for figure.
I will also need to adjust the change in omero-scripts

jburel:
@will-moore I have the changes to calculate the diff. I have also fixed the code in https://github.com/ome/omero-scripts/pull/214

will-moore:
That looks great. Wasn't sure it was worth it given that we don't *know* it's needed, but nice to be sure - thx

jburel:
Since the requirement has been modified i.e. minimum version for Pillow. Proposing to merge and tag as 5.18.0

sbesson:
Should https://github.com/ome/omero-py/blob/0958d2ba30f92990d6dea077fa359356577c6275/setup.py#L221 also be modified to capture the minimal Python requirement?

sbesson (approved):

PR #392, opened 2024-01-27
get objects by map annotations list

Josuerc026:
This pull request adds an additional gateway method, `getObjectsByMapAnnotationsList` to filter objects by a user provided list of map annotations, derived from the already provided `getObjectsByMapAnnotations`

```python
# List of key/value pairs to filter by
kv_pairs = [
    {
        'key': 'key1',
        'value': 'val1'
    },
    {
        'key': 'key2',
        'value': 'val2'
    },
    {
        'key': 'key3',
        'value': 'key3'
    }
]

# filter by object type and key/val filter
conn.getObjectsByMapAnnotationsList('Image', key_vals=kv_pairs)

#filter by object type, key/val filter, and namespace
conn.getObjectsByMapAnnotationsList('Image', key_vals=kv_pairs, ns='openmicroscopy.org/omero/client/mapAnnotation')
```

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -4293,7 +4293,7 @@ class _BlitzGateway (object):
             query += "and ann.ns = " + ns
         query += """
             group by obj.id 
-            having count(*) = %s
+            having count(*) >= %s
         """ % key_vals_count
 
         # Return objects
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -4290,7 +4290,7 @@ class _BlitzGateway (object):
         """ % (wrapper().OMERO_CLASS, keys_str, vals_str)
         # filter by namespace
         if ns is not None:
-            query += "and ann.ns = " + ns
+            query += "and ann.ns = '" + ns + "'"
         query += """
             group by obj.id 
             having count(*) >= %s
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -4280,16 +4280,21 @@ class _BlitzGateway (object):
             params.page(offset, limit)
 
         # Query
-        query = f"""
-            select distinct obj.id from {obj_type}AnnotationLink oal
+        query = """
+            select distinct obj.id from %sAnnotationLink oal
             join oal.parent obj 
             join oal.child ann 
             join ann.mapValue mv
-            where mv.name in {keys_str}
-            and mv.value in {vals_str}
+            where mv.name in %s
+            and mv.value in %s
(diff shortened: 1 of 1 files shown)


src/omero/gateway/__init__.py
```diff
@@ -4244,7 +4244,64 @@ def add_param(param, value):
...
+        
+        params = omero.sys.ParametersI()
+        # Total length of key/val pairs to match
+        # stores in a set to avoid duplicates
+        key_vals_count = len(key_vals)
+        keys = set()
+        vals = set()
+        for kv in key_vals:
+            keys.add(kv['key'])
+            vals.add(kv['value'])
```
will-moore, line 4269: Since you are adding both the keys and vals to different `sets` here, you lose the key-value pairing?

will-moore:
Hi @Josuerc026 and thanks for your proposed contribution.

 I don't have a clear idea of the use-case is for this, but I'm not sure that this functionality will be needed by enough users to justify it's inclusion in the BlitzGateway?

If you want to extend the BlitzGateway in your code to add this method to the class, you could adopt the approach we use in the webclient. See https://github.com/ome/omero-web/blob/9841167a70df2f0174aed2820bcd302873b53df3/omeroweb/webclient/webclient_gateway.py#L99

There we update the `omero.gateway.BlitzGateway` https://github.com/ome/omero-web/blob/9841167a70df2f0174aed2820bcd302873b53df3/omeroweb/webclient/webclient_gateway.py#L2101
so after importing the `webclient_gateway`, we can then import the BlitzGateway as usual and you will get the derived class with additional methods on it.

Josuerc026:
> Hi @Josuerc026 and thanks for your proposed contribution.
> 
> I don't have a clear idea of the use-case is for this, but I'm not sure that this functionality will be needed by enough users to justify it's inclusion in the BlitzGateway?
> 
> If you want to extend the BlitzGateway in your code to add this method to the class, you could adopt the approach we use in the webclient. See https://github.com/ome/omero-web/blob/9841167a70df2f0174aed2820bcd302873b53df3/omeroweb/webclient/webclient_gateway.py#L99
> 
> There we update the `omero.gateway.BlitzGateway` https://github.com/ome/omero-web/blob/9841167a70df2f0174aed2820bcd302873b53df3/omeroweb/webclient/webclient_gateway.py#L2101 so after importing the `webclient_gateway`, we can then import the BlitzGateway as usual and you will get the derived class with additional methods on it.

Hi @will-moore, no problem at all. I encountered a scenario where the above was helpful for my use case but I appreciate you taking the time to review it.

Feel free to close this out, and I will take a look at those resources you linked to 😄

PR #394, opened 2024-01-30
added `_repr_html_` method for `_ImageWrapper`

### LINKED ISSUE #393: Overwrite `_repr_html_` for `_ImageWrapper`
## Feature pitch: 🖼️ html metadata preview 🖼️ 

Hi all,

I have recently started to work a bit more with OMERO and, more specifically `ompero-py` and `ezomero`. I work predominatly with Jupyter notebooks and what I was often missing, was a clearer visualization of what I had actually just downloaded through the `omero` gateway. And I thought it would be cool if the `_ImageWrapper` object, specifically, had a bit more rich output into a Jupyter notebook than only `<_ImageWrapper id=xyz>`. In other libraries (e.g., dask, vedo, etc), just printing their respective objects shows some nice summaries of some properties of the objects into the notebook. This can be done by overloading the `_repr_html_` method of any class definition.

I started on a little implementation but I first wanted to ask whether it would actually be desirable for you to have this feature before going any further with this. In my current approach, retrieving an image like this:

```
metadata, data = ezomero.get_image(
    conn,
    image_id=some_id,
    no_pixels=True,
    dim_order='tczyx')
```

yields `metadata` and `data`, whereas `data` is `None` and `metadata` is of type `_ImageWrapper`. After overloading `_repr_html_`, simply typing `metadata` in a Jupyter notebook yields this overview:

Capture

In essence, it's a short html document, the values for which are filled with the `self.getName()`, `self.getId()`, etc member functions of the `_ImageWrapper` class. The functionality is only invoked in notebooks, to my knowledge, and should not touch any other functionality.

Looking forward to hearing your thoughts on this and send a PR if you are interested.

Thanks,
Johannes

jo-mueller:
Fixes #393 

## Description

This PR overwrites the `repr_html` for the `_imageWrapper` class to show a html overview object in Jupyter Notebooks whenever an `ImageWrapper` object is sown in a cell. In essence, it creates an html object and fills it with some essential information about the image. Currently displayed:

- Image ID
- Project ID
- Description
- Image Name
- Image dimensions (TCZYX)
- Voxel dimensions (currently output with long names (e.g. `MICROMETER` - there is probably a way to add the units as µm or whatever it is set to).
- Channel names

Let me know what else you would consider relevant or added/changed in the displayed object. I am by no means an html expert, so there is certainly some room for improvement there :)

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -10474,10 +10474,10 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
             {self.getSizeZ()},
             {self.getSizeY()},
             {self.getSizeX()})"""
-        physical_dims = f"""(
-            {self.getPixelSizeZ()},
-            {self.getPixelSizeY()},
-            {self.getPixelSizeX()})"""
+        physical_dims = """({:.3f}, {:.3f}, {:.3f})""".format(
+            self.getPixelSizeZ(),
+            self.getPixelSizeY(),
+            self.getPixelSizeX())
         physical_units = f"""(
             {self.getPixelSizeZ(units=True).getUnit()},
             {self.getPixelSizeY(units=True).getUnit()},
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -10426,6 +10426,107 @@ class _ImageWrapper (BlitzObjectWrapper, OmeroRestrictionWrapper):
             return count[0][0].getValue()
         return len(self._get_rois(shapeType, filterByCurrentUser))
 
+    def _repr_html_(self):
+        import base64
+
+        html_style_header = """
+        <!DOCTYPE html>
+        <html lang="en">
+        <head>
+            <meta charset="UTF-8">
+            <meta name="viewport" content="width=device-width, initial-scale=1.0">
+            <title>Image Details</title>
+            <style>
+                img {
+                    min-width: 250px; /* Set the minimum width for all images */
+                    min-height: 250px; /* Set the minimum height for all images */
+                }
+                .align-top {
+                vertical-align: top;
+                }
+                .text-right {
+                text-align: right;
+                }
+            </style>
+        </head>
+        """
+
+        # create a sub-table for image information
+        table_imageinfo = f"""
+        <table>\n
+            <tr>\n
+                <td><b>Image name: </b></td> <td class=text-right>{self.getName()}</td>\n
+            </tr>\n
+            <tr>\n
+                <td><b>Image ID: </b></td>  <td class=text-right>{self.getId()}</td>\n
+            </tr>\n
+            <tr>\n
+                <td><b>Project ID: </b></td>   <td class=text-right>{self.getProject().getId()}</td>\n
+            </tr>\n
+        </table>
+        """
+
+        # get entries for thumbnail and dimensions
+        encoded_image = base64.b64encode(self.getThumbnail()).decode('utf-8')
+        dimensions = f"""(
+            {self.getSizeT()},
+            {self.getSizeC()},
+            {self.getSizeZ()},
+            {self.getSizeY()},
+            {self.getSizeX()})"""
+        physical_dims = f"""(
+            {self.getPixelSizeZ()},
+            {self.getPixelSizeY()},
+            {self.getPixelSizeX()})"""
+        physical_units = f"""(
+            {self.getPixelSizeZ(units=True).getUnit()},
+            {self.getPixelSizeY(units=True).getUnit()},
+            {self.getPixelSizeX(units=True).getUnit()})"""
+
+        table_dimensions = f"""
+        <table>\n
+            <tr>\n
+                <td><b>Dimensions (TCZYX): </b></td> <td class=text-right>{dimensions}</td>\n
+            </tr>\n
+            <tr>\n
+                <td><b>Voxel/Pixel dimensions (ZYX): </b></td> <td class=text-right>{physical_dims}</td>\n
+            </tr>\n
+            <tr>\n
+                <td><b>Physical units: </b></td> <td class=text-right>{physical_units}</td>\n
+            </tr>\n
+            <tr>\n
+                <td><b>Channel Names: </b></td> <td class=text-right>{self.getChannelLabels()}</td>\n
+            </tr>\n
+        </table>
+        """
+
+        table_assembly = f"""
+        <table>
+        <tr>
+            <td><div class="thumbnail">
+                <img src="data:image/jpeg;base64,{encoded_image}" alt="Thumbnail">
+            </div></td>
+            <td class="align-top"><h2>Image information </h2>
+            {table_imageinfo}
+            </td>
+        </tr>
+        </table>
+        <table>
+            <tr>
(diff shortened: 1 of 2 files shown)


will-moore:
This looks good:

Screenshot 2024-01-31 at 15 51 51

You're missing Dataset info. You can get a Dataset with `self.getParent()` although it might return `None` if more than one (or orphaned). Also `getProject()` could return None so you need to handle that possibility.

I had a go with this...

```

        def obj_html(obj, otype):
            return f"""<tr>
                    <td><b>{otype}</b></td>
                    <td class=text-right>{obj.id if obj else ""}</td>
                    <td class=text-right>{obj.name if obj else ""}</td>
                    </tr>
                """

        # create a sub-table for image information
        table_imageinfo = f"""
        <table>
            <tr><th></th><th>ID</th><th>Name</th></tr>
            {obj_html(self, 'Image')}
            {obj_html(self.getParent(), 'Dataset')}
            {obj_html(self.getProject(), 'Project')}
        </table>
        """
```

Screenshot 2024-01-31 at 16 17 05

will-moore:
cc @pwalczysko @jburel - Any suggestions here?

pwalczysko:
> cc @pwalczysko @jburel - Any suggestions here?

What about pixel type and ROIs (number of ROIs) ?

jo-mueller:
@will-moore I think it'll make sense to reduce the precision on the displayed voxel dimensions. Should be an easy fix. I do like the idea about adding some basic information about the parent dataset

> Also getProject() could return None so you need to handle that possibility.

I think `str(None)` just returns `"None"`. It wouldn't exactly be pretty but I think a `None` entry in any of these values would at least not break the displayed `html`

> What about pixel type and ROIs (number of ROIs) ?

I thought about ROIs but eventually refrained from adding it to restrict myself a bit to the parameters that would *always* be set for an image. But then again, there is still plenty of space in the table(s) 🤔

src/omero/gateway/__init__.py
```diff
@@ -10426,6 +10426,107 @@ def getROICount(self, shapeType=None, filterByCurrentUser=False):
...
+        table_imageinfo = f"""
+        <table>\n
+            <tr>\n
+                <td><b>Image name: </b></td> <td class=text-right>{self.getName()}</td>\n
+            </tr>\n
+            <tr>\n
+                <td><b>Image ID: </b></td>  <td class=text-right>{self.getId()}</td>\n
+            </tr>\n
+            <tr>\n
+                <td><b>Project ID: </b></td>   <td class=text-right>{self.getProject().getId()}</td>\n
```
will-moore, line 10464: If `getProject()` returns `None` then this will fail with something like `Nonetype has no attribute getId`.
Same with `self.getProject().getName()`

will-moore:
Opened PR with those changes at https://github.com/jo-mueller/omero-py/pull/1

src/omero/gateway/__init__.py
jo-mueller: Ah, I get it! Will fix it asap.

jo-mueller:
Limited floating number precision makes it a bit prettier:

Capture

will-moore (approved):
Looks great, thanks

jo-mueller:
Cool! Let me know if I should change anything - I probably won't have so much time to work on it because I'll be on leave for a few months soon 👍

jo-mueller:
@will-moore @pwalczysko is there anything I should add to this PR to see it through? Happy for any feedback 👍

will-moore:
Hi @jo-mueller - apologies for dropping this.
I'd like to get a second opinion from someone on the OME team e.g. @jburel but in the meantime, it occurred to me that it might be nicer not to add all this html into the main `gateway.__init__.py` since it's already getting very verbose. Instead, the bulk of it could be moved to e.g. `omero.gateway.utils` as an `image_to_html(image)` method.

Also, in order to accept external contributions to any of the OME's GPL repos, we need to ask that you fill out a CLA form and submit it as described at https://ome-contributing.readthedocs.io/en/latest/cla.html

Thanks

will-moore:
Thanks for the CLA.

I would add the code to https://github.com/ome/omero-py/blob/master/src/omero/gateway/utils.py
e.g.

```
import base64

...

def image_repr_html(image):
    ...

   table_imageinfo = f"""
        <table>
            <tr><th></th><th>ID</th><th>Name</th></tr>
            {obj_html(image, 'Image')}
            {obj_html(image.getParent(), 'Dataset')}
            {obj_html(image.getProject(), 'Project')}
        </table>
        """
    ...

    return html
```

Then add `image_repr_html` to the imports at
https://github.com/ome/omero-py/blob/master/src/omero/gateway/__init__.py#L32

Then...
```
def _repr_html_(self):

    return image_repr_html(self)
```

jo-mueller:
Done in da9d6443549e8353ba3cd8c5aa4c952dca95620e

will-moore:
Final test with this branch installed locally, looking good...

Screenshot 2024-04-16 at 12 24 33

pwalczysko:
With an image which does have pixelsizes, I have success

Screenshot 2024-04-16 at 12 49 03



But with an image which does not have pixelsizes, I do get an error (image created by ``touch a.fake`` cmd):

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File ~/opt/anaconda3/envs/jotest/lib/python3.9/site-packages/IPython/core/formatters.py:344, in BaseFormatter.__call__(self, obj)
    342     method = get_real_method(obj, self.print_method)
    343     if method is not None:
--> 344         return method()
    345     return None
    346 else:

File ~/Work/omero-py/target/omero/gateway/__init__.py:299, in BlitzObjectWrapper._repr_html_(self)
    294 def _repr_html_(self):
    295     """
    296     Returns an HTML representation of the object. This is used by the
    297     IPython notebook to display the object in a cell.
    298     """
--> 299     return image_to_html(self)

File ~/Work/omero-py/target/omero/gateway/utils.py:276, in image_to_html(image)
    269 encoded_image = base64.b64encode(image.getThumbnail()).decode('utf-8')
    270 dimensions = f"""(
    271     {image.getSizeT()},
    272     {image.getSizeC()},
    273     {image.getSizeZ()},
    274     {image.getSizeY()},
    275     {image.getSizeX()})"""
--> 276 physical_dims = """({:.3f}, {:.3f}, {:.3f})""".format(
    277     image.getPixelSizeZ(),
    278     image.getPixelSizeY(),
    279     image.getPixelSizeX())
    280 physical_units = f"""(
    281     {image.getPixelSizeZ(units=True).getUnit()},
    282     {image.getPixelSizeY(units=True).getUnit()},
    283     {image.getPixelSizeX(units=True).getUnit()})"""
    285 table_dimensions = f"""
    286 <table>\n
    287     <tr>\n
   (...)
    299 </table>
    300 """

TypeError: unsupported format string passed to NoneType.__format__
```

pwalczysko (changes requested):
Please address the image with no pixelsizes problem ^^^ thank you.

jo-mueller:
Hi @pwalczysko , 

thanks for the review, good catch!

I added some code to escape this error if the properties haven't been set (b0a7fada46d8af0764eeb94882474a610fbadde4). In this case, I would now get this:

capture

pwalczysko (approved):

Issue #379, opened 2023-07-27
conda install issue


brendanofallon:
HI there, I'm running into a strange issue trying to install omero-py via conda on RHEL 9 and Ubuntu 20.04. But the install command recommended in the documentation never works, for instance: 
 
    $ conda install -c conda-forge omero-py

    Retrieving notices: ...working... done
    Collecting package metadata (current_repodata.json): done
    Solving environment: unsuccessful initial attempt using frozen solve. Retrying with flexible solve.
    Solving environment: unsuccessful attempt using repodata from current_repodata.json, retrying with next repodata source.
    Collecting package metadata (repodata.json): \ Killed

The OS always kills the conda install process midway through the install, and looks like the conda process is allocating way too much memory (just eyeballing memory usage on top, it increases rapidly while the command is running, until the OS finally just kills it).  This happens repeatedly, every time I try that command, on both distros above. I'm not really sure if this is an omero-py issue, but conda appears to function normally for other packages I've tried to installed. My conda version is 23.5.2

Notably, installing from the bioconda channel (`conda install -c bioconda omero-py`) works fine, except that version appears to be older and requires libssl1.0, which is no longer supported (or even available) on RHEL 9.
 Let me know if I can provide any more info

manics:
Are you using Python 3.11 by any chance? zeroc-ice 3.6 isn't available for 3.11 https://github.com/ome/omero-py/issues/360#issuecomment-1653971055
conda won't downgrade core packages like Python by default, instead it gets into a spiral of incompatible nested dependencies.

brendanofallon:
Hi, I'm using python 3.8.17 - but I just figured out that the omero-py from the 'ome' conda channel installs just fine with no issues (whew!). So to sum up: installing using the conda-forge channel runs into a weird conda issue and break, using the bioconda channel has a version which requires a pretty old libssl1.0 which isn't available on some newer linux distros,  but the one from the ome channel worked just fine with no issues. The documentation at https://omero.readthedocs.io/en/stable/users/cli/installation.html recommends conda-forge. So maybe this is just a matter of standardizing the packages from different channels and updating the docs

jburel:
Did you follow the notes from the README i.e.
```
conda create -n myenv -c ome python=3.8 zeroc-ice36-python
conda activate myenv
conda install -c conda-forge omero-py
```

DrKenHo-crick:
Using CenOS 7 - the recommend install above and on README just doesn't work.

[...]

PR #397, opened 2024-02-13
Update README to use -c conda-forge

joshmoore:
see https://github.com/ome/omero-py/issues/379#issuecomment-1936583831

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,9 +33,8 @@ to install it first, see `miniconda`_ for more details.
 
 To install ``omero-py`` using conda (preferred)::
 
-    conda create -n myenv -c ome python=3.8 zeroc-ice36-python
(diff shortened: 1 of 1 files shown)


will-moore (approved):
Works for me!

jburel:
Does it mean the correct version of ice is installed?
The reason we reverted to use the ome conda channel was because ice 3.7 bindings were installed when using conda-forge channel

joshmoore:
for me `-c ome` is no longer working because of arm, so I'm solely using `conda-forge` these days. I'm not sure how Ice 3.7 would be downloaded based on https://github.com/conda-forge/omero-py-feedstock/blob/main/recipe/meta.yaml#L35

jburel:
Things might have changed

PR #400, opened 2024-02-29
GetSession

jburel:
Logging in using ``joinSession`` could lead to ``_sessionUuid`` not being set.
This leads to error like 
```
omero.RemovedSessionException: exception ::omero::RemovedSessionException
{
serverStackTrace = ome.conditions.RemovedSessionException: No context for
at ome.services.sessions.state.SessionCache.getDataNullOrThrowOnTimeout(SessionCache.java:432)
at ome.services.sessions.state.SessionCache.getSessionContext(SessionCache.java:368)
```
See https://github.com/TheJacksonLaboratory/ezomero/issues/100 for background


sbesson (commented):
I think `omero-user-token` and `joinSession` is a red herring. I am able to reproduce the initial issue when calling `omero.gateway.BlitzGateway.getSession` using the following minimal example

```python
import omero
client=omero.client(server)
client.createSession(user, password)
with BlitzGateway(client_obj=client) as conn:
    conn.getUser()
    conn.getSession()
```

Minimally, the proposed change allows the session UUID to be retrieved dynamical and `conn.getSession()` works as expected in the scenario above.

This raises a few higher level API questions about using client objects directly which I could not answer by looking at the implementation or the documentation:

- are there other APIs depending on the internal `self._sessionUuid` value and if should this internal field be updated as well in `getSession()`?
- what is the expectation when passing a client object to the constructor of the BlitzGateway? Is the contract that the client should have an initialised session? Or is it possible to pass an empty client and use another API to create a session?
- related to the previous point, I was surprised to see that

  ```python
  import omero
  client=omero.client(server)
  client.createSession(user, password)
  with BlitzGateway(client_obj=client) as conn:
    conn.connect()
  ```
   returns `False` although this has some relationship to the fact that `_sessionUuid` is set to `None`. Is this a valid workflow we should consider or is it expected that the user should never call `connect()` if a client object is passed?

jburel:
I was thinking about similar issues last night and I agree that it requires some investigations

jburel:
From the ``_init_`` method
```
if self.c is None:
            self._resetOmeroClient()
        else:
            # if we already have client initialised, we can go ahead and create
            # our services.
            self._connected = True
            self._createProxies()
            self.SERVICE_OPTS = self.createServiceOptsDict()
```
This indicates that a connection is established so the client should have an established connection but surprisingly the ``_sessionUuid`` is not set

DavidStirling:
I'd also been looking into this and came to a similar conclusion.

Fetching the UUID in the `getSession()` function seems to fix this particular case, but I wonder if it may be worth instead checking for and setting `_sessionUuid` within the init function when a pre-existing client is passed in. This would also provide an opportunity to validate whether the client is actually set up/connected properly.

If my understanding is correct this would then allow users to join or rejoin the session by calling `conn.connect()` if anything goes wrong, which currently fails and/or tries to create a new session.

jburel:
Digging a bit more into the problem 
The reason that 
```
with BlitzGateway(client_obj=client) as conn:
  conn.connect()
```
returns ``False``
is due to the fact that information from the passed client are not taken into account
The mainly ``host`` when a new client is created (if ``session_id`` is ``None``
and username and password when the session is created

I use the following code snippet to validate
```
with BlitzGateway(client_obj=client, host=server) as conn:
    conn.setIdentity(user, password)
    print(conn.connect())
```

If ``self._sessionUuid`` is not ``none, then the new client does not need to created and the session is used to "rejoin"
setting the ``sessionUuid`` will solve one problem but it will not solve the assumption made in ``_resetOmeroClient`` about the host if a client with host is specified.

jburel:
Setting the ``sessionUuid`` from the client if available leads to other problems when 
testing via 
```
client=omero.client(server)
client.createSession(user, password)
with BlitzGateway(client_obj=client) as conn:
  conn.connect()
```

error

```
exception ::Glacier2::PermissionDeniedException
{
    reason = Internal error. Please contact your administrator:
 Wrapped Exception: (org.springframework.ldap.AuthenticationException):
[LDAP: error code 49 - INVALID_CREDENTIALS: Bind failed: ERR_229 Cannot authenticate user uid=admin,ou=system]; nested exception is javax.naming.AuthenticationException: [LDAP: error code 49 - INVALID_CREDENTIALS: Bind failed: ERR_229 Cannot authenticate user uid=admin,ou=system]
}
```

jburel:
```
client=omero.client(server)
client.createSession(user, password)
print(client.getSessionId())
c=omero.client(server)
c.joinSession(client.getSessionId())
print(c.getSessionId())
```
This works fine so something in the gateway is affecting that

jburel:
I found the problem:
if the ``sessionUuid`` is set, 
in the connect method 
* the client is deleted in ``_resetOmeroClient(self)`` so the session is no longer valid
* leading to the error when attempting to join that session

sbesson (commented):
This might have created some possible regressions one of the gateway integration tests - see https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/374/testReport/OmeroPy.test.integration.gatewaytest.test_chmod/TestDefaultSetup/testAuthorCanEdit/

Also it would be useful to update the PR title and description to match the latest changes and possibly list the different connection/reconnection scenarios that should be tested

jburel:
The regression is possibly due to the fact that the session is not discarded.
In that case the contract will have to be different
It seems that the "flexibility" of the constructor leads to the fact that some methods cannot be invoked depending on the constructor used

jburel:
I have updated the constructor with various scenarii

src/omero/gateway/__init__.py
```diff
@@ -1525,14 +1525,24 @@ def __init__(self, username=None, passwd=None, client_obj=None, group=None,
         self.ice_config = [os.path.abspath(str(x)) for x in [_f for _f in self.ice_config if _f]]
 
         self.host = host
+        if self.c is not None:
+            self.host = self.c.getProperty("omero.host")
```
DavidStirling, line 1529: I could imagine an scenario where a user tries to connect to a server with the wrong client object, which would then cause some ambiguity over exactly which server Blitz is connected to.

Could I suggest that we check if self.host is already set to something different here, and if there's a mismatch throw an error?

will-moore:
There is a test failing at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest.test_chmod/TestDefaultSetup/testAuthorCanEdit/
I can't be sure whether it is due to this PR - I haven't yet managed to get integration tests running on my local machine, so I can't test there. Since @jburel is away just now, he suggests I exclude to see if that fixes the test for now...

src/omero/gateway/__init__.py
jburel: We could add a check but i think people either specify the "client" or the host/user/password
Checking the host does not stop the connection to a wrong server cf. one example in the description. Using the "connect" specifying the sessionID could lead to that situation cf. one of the examples in the description

jburel: i think the gateway is too "flexible" leading to some very strange situations.

jburel: Should we also set the identity if a client with connection is specified?

DavidStirling: I agree that people *should* either specify client or host, but in the event that they don't I think it's safer to avoid silently overriding the user-supplied `host` parameter. Some of the autocompletion tools people now use do have a habit of trying to fill every argument. The user really should be reading the docs, but at the same time I don't think Gateway's Python docs are as helpful as they need to be.

You're probably right that Gateway's excessive flexibility here may be causing more problems than it solves. An alternative might be to validate the connection in the client object and outright reject calls with the other parameters specified - you either init with a connected client or supply credentials, no "mix and match".

jburel: There are other problems especially with the ``connect`` method since new clients get created (even if one was specified in the constructors.
So probably need to describe what we expect from each method moving forward

jburel: The more I look into it the more I notice points that require clarification.
The ``clone`` method for example will not work when a client with session is specified.
To make it works one should do 
```
BlitzGateway(client_obj=client)
conn.setIdentity(user, password)
```

src/omero/gateway/__init__.py
```diff
@@ -2029,7 +2049,7 @@ def isSecure(self):
         Returns 'True' if the underlying omero.clients.BaseClient is connected
         using SSL
         """
-        return hasattr(self.c, 'isSecure') and self.c.isSecure() or False
+        return hasattr(self.c, 'isSecure') and self.c.isSecure() or self.secure
```
joshmoore, line 2052: Based on line 1543 could this just be `return self.secure`?

src/omero/gateway/__init__.py
```diff
@@ -2082,8 +2102,12 @@ def _resetOmeroClient(self):
         logger.debug(self.ice_config)
 
         if self.c is not None:
-            self.c.__del__()
-            self.c = None
+            try:
+                if self.c.getSessionId() !=  self._sessionUuid:
```
joshmoore, line 2106: This surprises me a bit. If you've discovered things from the tests, etc., it might be good to have them here as comments. It my mind, this is trying to kill the client, keeping it alive could leave services, etc. active. If the problem is that some tests were passing this in but not expecting to have their clients killed, then `self.c.newClient()` might be an alternative solution.

src/omero/gateway/__init__.py
```diff
@@ -1525,14 +1525,33 @@ def __init__(self, username=None, passwd=None, client_obj=None, group=None,
...
         self.port = port
+        if self.c is not None:
+            pc = self.c.getProperty("omero.port")
+            if self.port is None:
+                self.port = pc
+            elif pc != self.port:
+                raise Exception("ports %s and %s do not match" % (pc, self.port))
         self.secure = secure
+        if self.c is not None:
+            self.secure = self.c.isSecure()
```
joshmoore, line 1543: Not overly important, but should there be a mismatch test here?

joshmoore (commented):
Few comments after a review triggered by the failure in https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/17/testReport/OmeroPy.test.integration.gatewaytest.test_chmod/TestDefaultSetup/testAuthorCanEdit/

src/omero/gateway/__init__.py
jburel: It could yes

src/omero/gateway/__init__.py
jburel: Without the change, the following
```
client=omero.client()
client.createSession(user, password)
with BlitzGateway(client_obj=client) as conn:
    conn.connect()
```
fails with 
```
 raise omero.ClientError("No session available")
```
This was unexpected

joshmoore: From [today's error](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/OmeroPy.test.integration.gatewaytest.test_connection/TestConnectionMethods/testSecure/):
```
>               self._sessionUuid = self.c.getSessionId()
E               AttributeError: '_BlitzGateway' object has no attribute 'getSessionId'
```

Does this mean that a gateway is getting passed in as `client_obj`?

jburel: Something in the test was incorrect, i passed the wrong object

jburel:
@DavidStirling I have added integration tests for the proposed changes.
Anything else you came across during your investigation?

src/omero/gateway/__init__.py
```diff
@@ -1936,6 +1955,7 @@ def close(self, hard=True):  # pragma: no cover
                 oldC = None
                 self.c = None
                 self._session = None
+                #self._sessionUuid = None
```
DavidStirling, line 1958: Is this meant to be commented out?

jburel: @DavidStirling sorted

DavidStirling:
LGTM, thanks

src/omero/gateway/__init__.py
jburel: tested added there https://github.com/ome/openmicroscopy/pull/6381/files

sbesson: My interpretation of @joshmoore's question was: what should happen if a gateway object is constructed with an initialized client object and `secure={True,False}` and the encryption state of the client as returned by `self.c.isSecure()` does not match the `secure` keyword?

src/omero/gateway/__init__.py
```diff
@@ -2029,7 +2049,7 @@ def isSecure(self):
         Returns 'True' if the underlying omero.clients.BaseClient is connected
         using SSL
         """
-        return hasattr(self.c, 'isSecure') and self.c.isSecure() or False
```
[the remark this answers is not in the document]
sbesson: The initial connection will (and should) always use an encrypted client. But  https://github.com/ome/omero-py/blob/4dc40d68a9e511c4dc81f718c514aeea15029a6e/src/omero/gateway/__init__.py#L2074 in `_createSession` should create an unencrypted client depending on the state of `secure`?

```python
>>> conn=BlitzGateway(username=user,passwd=password,host="localhost")
>>> conn.connect()
True
>>> conn.isSecure()
False
>>> conn.close()
>>> conn=BlitzGateway(username=user,passwd=password,host="localhost",secure=True)
>>> conn.connect()
True
>>> conn.isSecure()
True
```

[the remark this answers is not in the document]
jburel: i get that 
but as I said the client specified could be insecured and in that case even if ``secure=True`` is passed
``conn.isSecure()`` will be ``false``.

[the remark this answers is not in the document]
jburel: I have reverted the change and change it 
to 
```
def isSecure(self):
        """
        Returns 'True' if the underlying omero.clients.BaseClient is connected
        using SSL
        """
        return  hasattr(self.c, 'isSecure') and self.c.isSecure() or self.secure
```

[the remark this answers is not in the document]
sbesson: So the scenario we are discussing is 

```
conn=BlitzGateway(client_obj=insecure_client, secure=True)
conn=BlitzGateway(client_obj=secure_client, secure=False)
```

Interestingly, part of the changes proposed here are to throw an exception on the following use cases
```
conn=BlitzGateway(client_obj=client, host=different_host)
conn=BlitzGateway(client_obj=client, port=different_port)
```

My feeling would be to say `secure` should behave in the same way i.e. if the constructor receives conflicting information from its inputs, an exception should be sent back to the caller. Unlike `host` and `port` the default of `secure` is `False`. A possible (untested) solution would be to change the default to `None` to cater for this scenario (and internally default to `False` if no client object is passed).

sbesson (approved):
The last changes were tested with the following test script

```python
import omero

print("Connect with user/password, no secure argument")
with BlitzGateway(username=user, passwd=password, host=host) as conn:
    print("Connected: " + str(conn.connect()))
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")
    print("Creating encrypted client")
    conn.setSecure(True)
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")

print("\n")
print("Connect with user/password and secure=False")
with BlitzGateway(username=user, passwd=password, host=host, secure=False) as conn:
    print("Connected: " + str(conn.connect()))
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")
    print("Creating encrypted client")
    conn.setSecure(True)
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")

print("\n")
print("Connect with user/password and secure=True")
with BlitzGateway(username=user, passwd=password, host=host, secure=True) as conn:
    print("Connected: " + str(conn.connect()))
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")
    print("Creating unencrypted client")
    conn.setSecure(False)
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")

c = omero.client(host)
c.createSession(user, password)

print("\n")
print("Connect with encrypted client object")
with BlitzGateway(client_obj=c) as conn:
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")
    print("Creating unencrypted client")
    conn.setSecure(False)
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")
    c.closeSession()

c = omero.client(host)
c.createSession(user, password)
c2 = c.createClient(secure=False)
c.__del__()

print("\n")
print("Connect with unencrypted client object")
with BlitzGateway(client_obj=c2) as conn:
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")
    print("Creating encrypted client")
    conn.setSecure(True)
    print(f"conn.c.isSecure: {conn.c.isSecure()} "
          f"conn.isSecure: {conn.isSecure()} "
          f"conn.secure: {conn.secure}")
    c2.closeSession()

print("\n")
print("Connect with encrypted client object and secure=False")
c = omero.client(host)
c.createSession(user, password)
try:
    conn = BlitzGateway(client_obj=c, secure=False)
except Exception as e:
    print(str(e))
    c.closeSession()


c = omero.client(host)
c.createSession(user, password)
c2 = c.createClient(secure=False)
c.__del__()
print("Connect with unencrypted client object and secure=True")

try:
    conn = BlitzGateway(client_obj=c2, secure=True)
except Exception as e:
    print(str(e))
    c2.closeSession()
```

with the following output

```
Connect with user/password, no secure argument
Connected: True
conn.c.isSecure: False conn.isSecure: False conn.secure: False
Creating encrypted client
conn.c.isSecure: True conn.isSecure: True conn.secure: True


Connect with user/password and secure=False
Connected: True
conn.c.isSecure: False conn.isSecure: False conn.secure: False
Creating encrypted client
conn.c.isSecure: True conn.isSecure: True conn.secure: True


Connect with user/password and secure=True
Connected: True
conn.c.isSecure: True conn.isSecure: True conn.secure: True
Creating unencrypted client
conn.c.isSecure: False conn.isSecure: False conn.secure: False


Connect with encrypted client object
conn.c.isSecure: True conn.isSecure: True conn.secure: True
Creating unencrypted client
conn.c.isSecure: False conn.isSecure: False conn.secure: False


Connect with unencrypted client object
conn.c.isSecure: False conn.isSecure: False conn.secure: False
Creating encrypted client
conn.c.isSecure: True conn.isSecure: True conn.secure: True


Connect with encrypted client object and secure=False
Secure flag False and True do not match
Connect with unencrypted client object and secure=True
Secure flag True and False do not match
```

The state of the `secure` keyword is now consistent across all usages. Happy fo this to get merged assuming the integration tests pass

jburel:
Integration tests are green but I could add a few more to match your detailed review

jburel:
Supporting tests are now all green. Proposing to merge and release as 5.19.2 (alongside a few approved PRs)

sbesson:
> Supporting tests are now all green. Proposing to merge and release as 5.19.2 (alongside a few approved PRs)

No objection from my side. Do you have a shortlist of all PRs proposed for inclusion?

jburel:
short list: 
#394
#398
#406 
#405 (potentially but in that case tag will be 5.20.0)

will-moore:
Would be great to get all those PRs in 👍

sbesson:
Looks good to me. I am also in favor of including #405 as it reduces the possibility of calls which can cause memory issues server-side. As discussed on the PR, the public API is not modified so I think it could be amenable for either a patch or minor release increment.

PR #403, opened 2024-03-25
Consume pre-build Ice Python wheel packages on Windows builds

sbesson:
Uses the newly pre-built Ice wheel packages for Windows x86_64 - see https://github.com/glencoesoftware/zeroc-ice-py-win-x86_64/releases/tag/20240325

Builds and unit test should now be running for Python 3.8 -> 3.12 on Windows 2019

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -21,11 +21,10 @@ jobs:
           - '3.12'
         os:
           - ubuntu-22.04
+          - windows-2019
         include:
           - python-version: '3.8'
             os: macos-latest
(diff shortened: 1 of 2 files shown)


sbesson:
For a review of the impact of these pre-built packages on total execution time, review https://github.com/ome/omero-py/actions/runs/8419617008/usage (last build without this PR) vs https://github.com/ome/omero-py/actions/runs/8423913982/job/23066668892 (with this PR)

jburel (approved):
This is great
I can see us using in several other repositories

PR #406, opened 2024-04-18
update the build steps

jburel:
This matches the way the packages are built on merge-ci. This is now the recommended way

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -57,9 +57,9 @@ jobs:
           python-version: '3.9'
       - name: Build package
         run: |
-          python -mpip install wheel
-          python setup.py sdist bdist_wheel
+          python -mpip install build
+          python -m build
       - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@v1.3.0
+        uses: pypa/gh-action-pypi-publish@v1.8.14
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yml
```diff
@@ -57,9 +57,9 @@ jobs:
           python-version: '3.9'
       - name: Build package
         run: |
-          python -mpip install wheel
-          python setup.py sdist bdist_wheel
+          python -mpip install build
+          python -m build
       - name: Publish to PyPI
-        uses: pypa/gh-action-pypi-publish@v1.3.0
+        uses: pypa/gh-action-pypi-publish@v1.8.14
```
sbesson, line 63: Is this a requirement for the workflow/repository? Or just a regular maintenance upgrade?

sbesson (commented):
The changes are hard to test as they are only executed on release. The usage of the new build target is identical to https://github.com/ome/omero-scripts/actions/runs/8740926594 where similar changes were made and the last release was successful.

One question about the action bump which was not done in `omero-scripts`

.github/workflows/workflow.yml
jburel: regular maintenance

sbesson (approved):

PR #408, opened 2024-05-01
remove deprecated method

jburel:
Remove method

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -1916,42 +1916,6 @@ class _BlitzGateway (object):
             logger.debug("... error not reconnecting")
             return False
 
-    def seppuku(self, softclose=False):  # pragma: no cover
-        """
-        Terminates connection with killSession(). If softclose is False, the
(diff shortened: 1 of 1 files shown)


sbesson (commented):
This API has been deprecated since OMERO 5.3.0 and is certainly scheduled for removal. Is the plan to release this in a minor version increment or a major version increment given the backwards incompatible change? Also is there a concrete driver for actioning this removal?

jburel:
No real concrete driver but while working on the gateway I noticed the method marked as deprecated and the apologies linked to it so it does not make sense to keep it and I think it is more that time to remove it

sbesson (approved):
Agreed on the removal of a long-deprecated method. Probably the biggest question to answer is whether this should wait until a major version increment

sbesson:
Discussed at the OMERO PR review. No objection with the spirit of the removal but this is effectively a backwards incompatible API change. Rather than creating a precedent, closing this and we can remove once we decide on working towards OMERO.py 6.0.0

PR #410, opened 2024-05-29
use gs wheel to match doc

jburel:
This PR modifies the wheels used for testing when using Python 3.8, 3.9 and 3.10
This is to match the recommendation introduced in https://github.com/ome/omero-documentation/pull/2427


sbesson (approved):
All good from my side

will-moore:
LGTM 👍

PR #411, opened 2024-05-29
Update readme

jburel:
Update the installation instructions 
The readthedocs page will be updated when a new tag is pushed. 
I will do that later today


will-moore (approved):
looks good, thx

PR #416, opened 2024-06-19
Parse omero.logging properties under new Logging heading

sbesson:
See https://github.com/ome/openmicroscopy/pull/6393, this adds the corresponding changes to the `omero config parse [--rst]` command so that the properties gets properly parsed and included in a new section under https://omero.readthedocs.io/en/stable/sysadmins/config.html

To test these changes, a server binary needs to be built with the changes from https://github.com/ome/openmicroscopy/pull/6393. Running `OMERODIR=path_to_server omero config parse --rst` should now include a Logging section with all `omero.logging` properties descriptions and their default values.

I expect that with the current 5.6.11 release of OMERO.server which does not includes these properties under `etc/omero.properties`, these additions should be a no-op and not impact the generated documentation.

### DIFF
--- a/src/omero/install/config_parser.py
+++ b/src/omero/install/config_parser.py
@@ -44,6 +44,7 @@ GRID_HEADER = Header("Grid", reference="grid")
 ICE_HEADER = Header("Ice", reference="ice")
 LDAP_HEADER = Header("LDAP", reference="ldap")
 JVM_HEADER = Header("JVM", reference="jvm")
+LOGGING_HEADER = Header("Logging", reference="log")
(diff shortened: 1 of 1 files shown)


will-moore:
Does this PR purely affect the `omero config parse` command (just for documentation/info) or does it have some other functional effect?

sbesson:
To the best of my knowledge, the only place where this codepath is used is the `omero config parse` command. It is certainly exclusively intended for documentation purposes.

joshmoore (approved):
:+1:

PR #419, opened 2024-07-05
admin rewrite: make OMERO.tables module configurable

sbesson:
Add @omero.tables.module@ to the list of templates to be substituted in the server IceGrid XML templates and replaces it with the value. of `omero.tables.module` in the OMERO configuration, using `runTables` as the default.

This change should be a no-op when used with OMERO.server binaries up to 5.6.11 but will be mandatory once the templates are updated to use the new substitution field.


chris-allan (approved):

sbesson:
0d8fc4a is similar to https://github.com/ome/omero-py/pull/416 and allows `omero config parse` to detect all properties starting with `omero.tables` and list them under a new `Tables` header in the [Configuration property glossay](https://omero.readthedocs.io/en/stable/sysadmins/config.html)

PR #427, opened 2024-08-30
Add fileset folder creation

### LINKED ISSUE #425: conn.c.download behavior differs from java implementation
Hello, 

I noticed that the behavior of the [conn.c.download](https://github.com/ome/omero-py/blob/749a68acfc7b1851c165ae75f2c118743d8b1e0d/src/omero/clients.py#L901) method is not the same as the one implemented on omero.inisght (and [omero-java-gateway](https://github.com/ome/omero-gateway-java/blob/62d4ba9e6cab7d6a7bdd1ab5d0c72695e4f71fea/src/main/java/omero/gateway/facility/TransferFacilityHelper.java#L85))
In the latest version of omero.inisght, when an image is downloaded, a parent folder `Fileset_xxxx` is created and the image(s) linked to the corresponding fileset are downloaded inside.

This is not the case in omero-py. The image is simply downloaded without creating any folder.

I was wondering if it is possible to mirror the behavior of the latest java implementatio into omero-py. It is quite annoying to deal with those two behaviors in automated tasks, espacially when I have to read the images froma a different script.

Thanks,
Rémy.

Edit : current existing bug in Java implementation for vsi files : https://github.com/ome/omero-gateway-java/issues/89

Rdornier:
Fix #425 by adding the fileset layer in the download folders

### DIFF
--- a/src/omero/plugins/download.py
+++ b/src/omero/plugins/download.py
@@ -90,6 +90,7 @@ class DownloadControl(BaseControl):
     def download_fileset(self, conn, fileset, dir_path):
         self.ctx.out(f"Fileset: {fileset.id}")
         template_prefix = fileset.getTemplatePrefix()
+        dir_path = os.path.join(dir_path, f"Fileset_{fileset.id}")
         for orig_file in fileset.listFiles():
             file_path = orig_file.path.replace(template_prefix, "")
             target_dir = os.path.join(dir_path, file_path)
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@ build
 dist/
 target/
 .DS_Store
+.idea
 _build


sbesson (changes requested):
I don't know if @will-moore or @jburel had a chance to look at this yet but from my side the primary issue with the current proposal is that is it a backwards incompatible API change as the default behavior of the `download_fileset` API and of the `omero download` command is altered.

In order to enable the new functionality in a backwards-compatibility manner, could a new keyword be added to `download_fileset` allowing to toggle the creation of the intermediate fileset folder?

sbesson:
Thanks @Rdornier, the new option makes sense from a code review perspective and the unit tests are passing. I'll pass onto @jburel and @will-moore for inclusion in the nightly CI builds (although please note the Dundee team is also working on CI migration so this might wait a few days until the infrastructure stabilizes).

will-moore:
@Rdornier did you need to make any other changes here, such as adding `insert_fileset_folder` argument to the code above where this is called, and maybe to the `$ omero download Fileset:1 ` command?

Or are you calling this function directly from elsewhere? You're using a script based on mine at
https://gist.github.com/will-moore/a9f90c97b5b6f1a0da277a5179d62c5a right?

src/omero/plugins/download.py
```diff
@@ -61,6 +59,8 @@ def _configure(self, parser):
             "OriginalFile is assumed if <object>: is omitted.")
         parser.add_argument(
             "filename", help="Local filename (or path for Fileset) to be saved to. '-' for stdout")
+        parser.add_argument(
+            "insert_fileset_folder", help="Adding 'Fileset_xxxx' folder in the download path", default="False")
```
will-moore, line 63: I think here you want to use the `"--insert_fileset_folder" action="store_true"` instead of `default="False"` 
Then you can check with simply `if args.insert_fileset_folder:`

e.g. see examples at
https://github.com/ome/omero-py/blob/4fd72efaa8cdbec9da98e1d28678af16878a86ba/src/omero/plugins/admin.py#L269

src/omero/plugins/download.py
```diff
@@ -72,24 +72,31 @@ def __call__(self, args):
         conn = BlitzGateway(client_obj=client)
         conn.SERVICE_OPTS.setOmeroGroup(-1)
 
+        if args.insert_fileset_folder:
+            insert_fileset_folder = True
+        else:
+            insert_fileset_folder = False
```
will-moore, line 78: Code is working fine. 
These 4 lines would be neater as simply `insert_fileset_folder = args.insert_fileset_folder` but not a blocker.

Rdornier: Done

will-moore (approved):
Great, thanks

sbesson (approved):

Issue #438, opened 2024-12-10
Unknown debug action: {'9'}


tstoeter:
Invoking `omero -d9 login idr.openmicroscopy.org` outputs
```
Unknown debug action: {'9'}
```
although the help `omero help debug` lists it as a valid option:
```
...
Available debugging choices:

    '0'         Disable debugging
    'debug'     Enable debugging at the first debug level
    '1'-'9'     Enable debugging at the specified debug level
    'trace'     Run the command with tracing enabled
    'profile'   Run the command with profiling enabled
...
```

PR #439, opened 2024-12-10
Fixes issue #438

tstoeter:
Python range does not include stop position. In this case it is 9, but should be 10.


will-moore (approved):
Thanks.
Fix looks good and corresponds to the help docstring 👍 

However, briefly looking at the usage of debug level I don't see that you'll get much more detailed output with level 9 compared with 8.

joshmoore:
Re-open to make use of #440

joshmoore:
Thanks, @tstoeter. Good point, @will-moore. I assume it's not overly critical though to reduce that now.

Issue #436, opened 2024-11-27
image_to_html always fails to find pixelSize


Tom-TBT:
An example of the issue from image.sc
https://forum.image.sc/t/fetch-and-save-ome-tiff-from-omero-to-local-directory-with-metadata/105507/3

It seems to me that something changed about `image.getPixelSizeX()` (and Y and Z)

With older versions of omero-py I was used to get this output:
```
object #0 (::omero::model::Length)
{
    _value = 0.16249999999999787
    _unit = MICROMETER
}
```

but it now only returns a float, so `0.16249999999999787` in this case.
What happens with `image_to_html` is that it tries to get the unit of the value from the returned pixel size. But since we get a float, it throws an error. The error is caught and the output defaults to `(na, na, an)`

It's still possible to get the old output by specifying the pixel units: `image.getPixelSizeX(units="MICROMETER")`. 

In the case of `image_to_html`, it would thus need to find the most appropriate unit to express the pixel size with as few 0s as possible.

I had a look with git blame to https://github.com/ome/omero-py/blame/master/src/omero/gateway/__init__.py#L9997 and https://github.com/ome/omero-py/blame/master/src/omero/gateway/__init__.py#L292 but everything is 10yo. 
I don't know what happened in between.

@jo-mueller

sbesson:
@Tom-TBT I tried to reproduce the issue using the following example code:

```python
from omero.gateway import BlitzGateway

with BlitzGateway(
        host='localhost', username=<username>, passwd=<password>) as conn:
    conn.connect()
    conn.SERVICE_OPTS['omero.group']='-1'

    i=conn.getObject('Image',185001)
    print(i.getPixelSizeX())
    print(i.getPixelSizeX(units="MICROMETER"))
```

All OMERO.py versions between 5.13.0 and 5.19.5 produced the following output:

```
0.5933762649494021
0.5933762649494021 MICROMETER
```

Older versions failed to connect due to SSL issues.

Matching your code history review, it looks like this has been the behavior of the API for the last decade. Do you have more specifics on which version of OMERO.py would have been returning the Length object by default? Or is the issue not about the `getPixelSize` API but rather the `image_to_html`?

Tom-TBT:
Hi @sbesson, thank you for the thorough testing.

> It seems to me that something changed about image.getPixelSizeX() (and Y and Z)

I was wrong, nothing changed then. As you say it's just about `image_to_html`

[...]

PR #429, opened 2024-09-06
move `repr_html` to `_ImageWrapper`

jo-mueller:
Fixes #428 

Dear OME-Team,

this PR fixes the issue that some classes that derive from `BlitzObjectWrapper` do not possess the necessary attributes to display the `repr_html` method correctly. `repr_html` mthods should be introduced specifically for each derived object. In this PR I simply move the `repr_html` to the derived class.

Since this is my last day at work before a longer parental leave, I unfortunately don't have the time to test this in detail but I hope that it's enough to get you started on the fix.

PR #441, opened 2024-12-19
Fix image_to_html pixel unit

Tom-TBT:
Fixes https://github.com/ome/omero-py/issues/436

The PR https://github.com/ome/omero-py/pull/394 introduced html representation of images, used for example when using jupyter notebooks.

An error slipped in with getting the pixelUnit of an image, this is simply the fix.

Before the fix

image

After the fix

image


sbesson:
Thanks for the review, thoughts about merging this alongside #429 and releasing 5.19.6?

PR #443, opened 2025-01-30
Gateway tests: replace DV files download by synthetic test images

sbesson:
The OMERO.py gateway tests have historically been relying on downloading [public Deltavision samples](https://downloads.openmicroscopy.org/images/gateway_tests/) for its Pytest fixtures. Recent connections issues between [OME CI](http://ci.openmicroscopy.org/) and [OME downloads](https://downloads.openmicroscopy.org/) have exposed the fragility of this set-up.

This PR updates this set-up to remove the dependency on a proprietary file formats as well as external data downloads and use Bio-Formats to generate synthetic test images with the desired features as described [here](https://bio-formats.readthedocs.io/en/stable/developers/generating-test-images.html) to .  

Four synthetic files are introduced for creating fixtures:
- `testimg.fake` a multi-Z/multi-C image with instrument, channel and plane metadata
- `tinyimg.fake` a minimal multi-Z/multi-T image with channel metadata
- `bigimg.fake` a large X/Y 8 bit image with multiple resolution levels
- `32float.fake` a large X/Y 32-bit floating point image 

The instrument and channel metadata depend on improvements introduced in [Bio-Formats 8.1.1](https://snoopycrimecop.github.io/www.openmicroscopy.org/2025/03/05/bio-formats-8-1-1.html) so the integration tests will require a version of OMERO.server with a recent Bio-Formats version.


will-moore:
Sounds like a good idea. Could we bundle a tiny text ome.xml image into the repo to test Channel wavelengths? Or maybe just skip it if no other option?

jburel:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/297 3 failures this morning

sbesson:
In addition to the 2 tests I reported initially which I was expecting to fail after my last changes,  `OmeroPy.test.integration.gatewaytest.test_wrapper.TestWrapper.testAllObjectsWrapped` is also attempting to retrieve metadata (an instrument) which does not exist in these fake files.

I could use some guidance on how you want to proceed with next steps. Translating the  DV into OME-XML as suggested in https://github.com/ome/omero-py/pull/443#issuecomment-2625658138 is an interesting idea as it would preserve the metadata. I had initially tried this route but realised these OME-XML files must be packaged as part of `omero-py`. Is this something you would consider and if so, should these be addedunder `omero/gateway/scripts` or elsewhere?

jburel:
Maybe we could consider adding some sample file from downloads.o.org? This means that the tests could not be run offline
Also we may want to review the content of https://github.com/ome/omero-common/tree/master/src/test/resources/ome

jburel:
Adding a test resources similar to the one in omero-common could be an option but that means that we will have 2 packages providing test file
How much work will it be to expand fake reader?

jburel:
Excluding this PR until a solution is implemented
--exclude

jburel:
This is in the same spirit that what has been done in https://github.com/ome/openmicroscopy/blob/develop/components/tools/OmeroJava/test/integration/ModelMockFactory.java
Another could be to handle an "xml" key in the name of the fake file
i.e. ``testimg&xml=BlobOfOMEXML.fake``
This could eventually replace the various existing conditions but that will be a broken change and probably not worth the effort

sbesson:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastCompletedBuild/testReport/ 

```
failed on setup with "OSError: [Errno 36] File name too long
```

I'll convert these into INI files and rework the logic

sbesson:
After a bit of investigation, I decided to store the OME-XML metadata of the public sample DV files used for the gateway tests under `src/omero/gateway/scripts/imgs`. Together with the logic in https://github.com/ome/omero-py/blob/05d52412ff7d6c4e3a515717d2807b1a21bf4291/setup.py#L202, this files should be included in the package and can be used for import by the integration tests.

This means that this PR no longer depend on the new feature introduced in https://github.com/ome/bioformats/pull/4272.

PR #444, opened 2025-02-25
omero.util.populate_metadata: only log column types from HeaderResolver

sbesson:
This should address a code scanning alert about potential logging sensitive information

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,48 @@
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches: [ "develop" ]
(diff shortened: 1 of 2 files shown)


sbesson:
Would it be preferable to have [4dbfb0b](https://github.com/ome/omero-py/pull/444/commits/4dbfb0b8633085571fc5317ee6189bbe39e3c8d4) as a separate PR (or pushed to `origin/master`) ?

jburel:
I think it is fine as it,
Merging so we can validate the new workflow

jburel:
https://github.com/ome/omero-py/actions/runs/13892309408 The build failed

PR #445, opened 2025-02-28
Relax numpy<2 capping

sbesson:
First step towards the resolution of https://github.com/ome/openmicroscopy/issues/6392

Like other projects, omero-py took the conservative approach of capping the numpy version following the release of 2.0.0. More than 6 months later, most of the ecosystem is supporting numpy 2 and even starts to require it. This PR updates OMERO.py to be compatible with Numpy 2.

Given the incompatibilities on Python 3.9 and the absence of compatible version of Numpy 2.x and PyTables, the proposal is as follows:
- for Python 3.9, keep the capping to `numpy<2` as currently
- for Python 3.10 and later, relax the capping to `numpy<3`

Existing OMERO.py installations on Python 3.10+ can elect to upgrade `numpy` to 2.x depending on their requirements.

Future steps outside this PR will include:
- removal of the Python 3.9 special casing when this version comes end of life in October 2025
- enforcement of `numpy` 2.x


sbesson:
The failures in [1a1599c](https://github.com/ome/omero-py/pull/445/commits/1a1599cd551d0b391c9e6fbfe74d6e6c8bd0a453) are related to the fact [PyTables 3.10.0](https://github.com/PyTables/PyTables/releases/tag/v3.10.0) introduced support for Numpy 2 but also requires Python 3.10.0. In other terms, there is currently no supported combination of Python 3.9/PyTables/Numpy 2.

[1b8c6f6](https://github.com/ome/omero-py/pull/445/commits/1b8c6f6bef2c3eb802692d41cb4f716af5644d10) is one way to address this issue i.e. uncap `numpy` for Python 3.10+ and keep it capped for Python 3.9. This special casing can be be dropped when support for Python 3.9, which becomes EOL in October 2025, is removed.

Note `tables` issue is only a server-side OMERO.py dependency required to start the PyTables-based OMERO.table service. A separate approach would be to extract this logic into a separate package that would only be installed server-side and uncap `numpy` for all Python versions on `omero-py`. Doing this would require scoping and discussions given the impact on deployment/upgrades.

This approach also raises questions around the testing environment. While the unit tests are executed against all supported Python versions, integration tests are currently performed within the OME CI infrastructure which is based on Rocky Linux 9 and uses the default `python3` package, i.e. Python 3.9. Without changes to the infrastructure and/or the packaging, there is no easy way to run the nightly CI OMERO integration tests with Numpy 2.

As a proposed next step, I will deploy a test OMERO.server with Numpy 2 and run the integration tests manually to have a first feeling of how many might be affected by the numpy upgrade either client side or server side. We can probably discuss next steps in terms of version support commitment, packaging and testing infrastructure.

jburel:
I think we should include this PR and release, This is creating many issues when using the library in Google Colab which already shipped with Numpy > 2
```
Uninstalling numpy-2.0.2:
      Successfully uninstalled numpy-2.0.2
```

DavidStirling:
Did some testing with other tooling using `numpy==2.2.4` on Python 3.10, 3.11 and 3.12. The lack of 3.13 ice builds for intel Macs prevented me from going further, but so far...

- omero2pandas seems absolutely fine from the client end, no issues with table read/write
- Segmentation tooling had no problems, we'd actually been using numpy 2 for a while without issue (older omero-py builds didn't pin)
- GS CellProfiler builds are pinned to numpy 1 by other dependencies, though that project is also held back on Python 3.9 by `wxpython` so I wouldn't consider this a major issue for now
- omero-web doesn't seem to mind, judging from a brief tour around the UI. I imagine it'd be worth following the full testing script (if we have one established for this sort of thing)

Running the `ruff` compatibility checker (`NPY201`) didn't find any issues in key repos such as omero-py and omero-web. In contrast it correctly found that CellProfiler is incompatible without substantial work (as expected). I think that means we're in reasonably good shape on this.

sbesson:
As discussed in [last week's OME meeting](https://www.openmicroscopy.org/minutes/2025/2025-04-01.html), the current plan is for @jburel to run a last series of checks in a Google Collab notebook environment.

jburel (approved):

Issue #356, opened 2022-12-06
TypeError: catching classes that do not inherit from BaseException is not allowed


will-moore:
Since at https://github.com/ome/omero-py/blob/ccfdadd131bf0f1bb4c4ac902a5c2aa8cbc712f5/src/omero/gateway/__init__.py#L7949 the `ctx.ignoreExceptions` is None by default, we effectively get:

```
try:
...
except None:
```
which gives following stack trace:

```
Traceback (most recent call last):

File "/opt/omero/web/venv3/lib64/python3.6/site-packages/django/core/handlers/exception.py", line 47, in inner

response = get_response(request)

File "/opt/omero/web/venv3/lib64/python3.6/site-packages/django/core/handlers/base.py", line 181, in _get_response

response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omeroweb/decorators.py", line 544, in wrapped

retval = f(request, *args, **kwargs)

File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omeroweb/webgateway/views.py", line 1076, in render_image

jpeg_data = img.renderJpeg(z, t, compression=compress_quality)

File "/opt/omero/web/venv3/lib64/python3.6/site-packages/omero/gateway/__init__.py", line 7949, in wrapped

except ctx.ignoreExceptions:

TypeError: catching classes that do not inherit from BaseException is not allowed
```

Issue #308, opened 2021-10-21
assert_re ignoreExceptions is None


will-moore:
If an exception is thrown during init of rendering engine, this throws an Exception which then also generates a `TypeError` with `except ctx.ignoreExceptions:` since `ignoreExceptions` is None by default 
see https://www.openmicroscopy.org/qa2/qa/feedback/30554/

```
Traceback (most recent call last):

File "/omero/omerowebenv/lib/python3.6/site-packages/omero/gateway/__init__.py", line 7941, in wrapped
if not self._prepareRenderingEngine() \

File "/omero/omerowebenv/lib/python3.6/site-packages/omero/gateway/__init__.py", line 8217, in _prepareRenderingEngine
self._re = self._prepareRE(rdid=rdid)

File "/omero/omerowebenv/lib/python3.6/site-packages/omero/gateway/__init__.py", line 8191, in _prepareRE
re.resetDefaultSettings(True, ctx)

File "/omero/omerowebenv/lib/python3.6/site-packages/omero/gateway/__init__.py", line 4856, in __call__
return self.handle_exception(e, *args, **kwargs)

File "/omero/omerowebenv/lib/python3.6/site-packages/omeroweb/webclient/webclient_gateway.py", line 2109, in handle_exception
super(OmeroWebSafeCallWrapper, self).handle_exception(e, *args, **kwargs)

File "/omero/omerowebenv/lib/python3.6/site-packages/omero/gateway/__init__.py", line 4853, in __call__
return self.f(*args, **kwargs)

File "/omero/omerowebenv/lib/python3.6/site-packages/omero_api_RenderingEngine_ice.py", line 2323, in resetDefaultSettings
return _M_omero.api.RenderingEngine._op_resetDefaultSettings.invoke(self, ((save, ), _ctx))

omero.LockTimeout: exception ::omero::LockTimeout
{
serverStackTrace = ome.conditions.LockTimeout: Pixels pyramid missing, being created or import in progress.

...

at java.base/java.lang.Thread.run(Thread.java:829)

serverExceptionClass = ome.conditions.LockTimeout
message = Pixels pyramid missing, being created or import in progress.
backOff = 15000
seconds = 0
}

During handling of the above exception, another exception occurred:

Traceback (most recent call last):

File "/omero/omerowebenv/lib/python3.6/site-packages/django/core/handlers/exception.py", line 41, in inner
response = get_response(request)

[...]

PR #447, opened 2025-03-04
Change ignoreExeptions default to empty tuple

### LINKED ISSUE #446: omero.gateway assert_re decorator raises TypeError during exception checking
https://github.com/ome/omero-py/blob/05d52412ff7d6c4e3a515717d2807b1a21bf4291/src/omero/gateway/__init__.py#L7927
https://github.com/ome/omero-py/blob/05d52412ff7d6c4e3a515717d2807b1a21bf4291/src/omero/gateway/__init__.py#L7958

The default value of ignoreExceptions in the assert_re decorator initializer is None but `except None` is invalid, raising a TypeError:
```
TypeError: catching classes that do not inherit from BaseException is not allowed
```
Exception chaining does retain the original exception and traceback but it's confusing to see this extra unrelated error.

`except None` was allowed in Python 2, where the argument to `except` could be anything and non-exception types would simply never match any actual raised exception. The no-match equivalent for Python 3 is an empty tuple, so I think the default arg for ignoreExceptions should be changed to `()`.

jmuhlich:
This is a Python 3 compatible way to specify that the except statement below should not catch anything, as intended by the previous default of None.

Fixes #446 #356 #308

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -7924,7 +7924,7 @@ class assert_re (object):
     call. Is configurable by various options.
     """
 
-    def __init__(self, onPrepareFailureReturnNone=True, ignoreExceptions=None):
+    def __init__(self, onPrepareFailureReturnNone=True, ignoreExceptions=()):
         """
         Initialises the decorator.
 


will-moore:
Great, thanks for the PR, we'll give it a test...

will-moore (approved):
Testing locally, manually introducing errors, this fix looks good and avoids the reported
```
TypeError: catching classes that do not inherit from BaseException is not allowed
```

sbesson (changes requested):
Thanks @jmuhlich. Can you also update the default value in the docstring of the method?

PR #450, opened 2025-03-12
Add model stubs

### LINKED ISSUE #448: Type annotations for `models`
Currently `omero.models` is a magic dynamic module whose contents are only created at runtime, as far as I know. This is annoying for two reasons. 

The first is that my type checker says I'm doing something wrong when I access `omero.model.DatasetI`, and the second is that the type checker doesn't know or autocomplete `omero.model.DatasetI.field`.

There are tools for generating type stubs in these scenarios. For example after installing `mypy` you can run `uv run `stubgen -m omero.model --inspect-mode` which outputs a massive file, a snippet of which is:

```python
class DatasetI:
    def getName(self, current): ...
    def setName(self, _name, current, wrap: bool = ...): ...
```

The actual types are missing which is a bit annoying, but this would still be a massive improvement.

multimeric:
Closes #448.


joshmoore:
Thanks for opening this, @multimeric!

I'm failing to get the `mypy` errors to go away. Could you perhaps provide an example of what you're seeing before and after the change?

multimeric:
Can you attach some of these errors please? The PR won't solve all mypy errors but it should eliminate some and clarify others.

multimeric:
Hmm part of the problem is these top level files that live directly in `site-packages`, like `omero_model_CodomainMapContextI.py`, which are used directly in the unit tests instead of `omero.model`. They make it incredibly difficult to install `.pyi` files and even if you do it, it doesn't seem like mypy detects them. Is there any plan to move them into a module?

joshmoore:
There's no current plan, no, but happy to either help someone do it or review work to attempt it.

sbesson:
Closing as discussed in the weekly OMERO PR review meeting. In addition the maintenance cost of adding 21K lines of code, we have no internal driver to evaluating these changes. As discussed above, rearchitecting `omero-py` to move the model classes under namespace might be a prerequisite in order to resurrect this but this will be a backwards-incompatible change which is currently unplanned.

PR #458, opened 2025-04-03
omero admin jvmcfg: remove MaxPermSize option

### LINKED ISSUE #61: Activate admin unit tests
see: https://github.com/ome/omero-py/pull/48

By downloading the contents of OMERO dist's `etc/` directory, the unit tests could be made to run.

sbesson:
Fixes #61. See also https://github.com/ome/openmicroscopy/pull/6428.

`-XX:MaxPermSize` has been ignored since Java 8. This PR updates the `omero admin` logic to remove this flag. The `-XX:+IgnoreUnrecognizedVMOptions` flag is also removed although this could be discussed as part of the review and the flag could be reintroduced conservatively.  The JVM configuration unit tests are adjusted to reflect these changes.

Additionally, the CI logic is updated to download the latest OMERO.server and  set `OMERODIR` on Ubuntu runners. The Tox configuration is also updated to pass `OMERODIR` through the tests. This should have the impact of restoring the execution of all unit tests requiring `OMERODIR` (primarily admin unit tests) on Ubuntu runners which is the primary environment where these commands would be executed. This  main downside to this coverage addition is a longer execution time for the Ubuntu jobs.

In terms of testing
- in GitHub CI, review the output of the unit tests especially on Ubuntu runners. The number of skipped tests (currently 129 in the latest execution against the HEAD) should decrease substantially and all the `test_jvmcfg.py` tests in particular should be executed and successful 
- in an environment with OMERO.server and OMERO.py including this PR, `omero admin jvmcfg` should only list the `-Xmx` option with a value depending on the configuration/system memory. The server should be able to restart with issue and all services should remain functional.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -36,6 +36,13 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Download the latest OMERO.server
+        if: matrix.os == 'ubuntu-22.04'
+        run: |
+          curl -L -o server.zip https://downloads.openmicroscopy.org/latest/omero/server.zip
+          unzip server.zip
+          ln -s OMERO.server* OMERO.current
+          echo "OMERODIR=${{ github.workspace }}/OMERO.current" >> $GITHUB_ENV
       - name: Install dependencies
         run: python -mpip install --upgrade wheel pytest tox virtualenv
       - name: Run tests
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Download the latest OMERO.server
         if: matrix.os == 'ubuntu-22.04'
         run: |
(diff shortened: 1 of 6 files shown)


joshmoore (commented):
Generally all makes sense. :100: for the cleanup.
 * For `-XX:+IgnoreUnrecognizedVMOptions`, I could see ways we could make this an env/cli option that could be activated for each of use in case something fails in the future, but probably not overly important. 
 * :+1: for downloading from GitHub, the `gh` tool which is installed in should be able to do this without the hard-coding (unless the hard-coding is what you want): `gh release download -R ome/openmicroscopy --pattern "OMERO.server*zip"`

sbesson:
Thanks for the GitHub CLI command suggestion, this addresses the only concern I had about this strategy and how to keep up with the latest server release.

joshmoore (approved):

sbesson:
This has been successfully included in a few builds. Given the review above, I would be inclined to get this in and possibly cut a pre-release of OMERO.py 5.20.0rc1 with Numpy 2.x support. @jburel would that make sense?

PR #460, opened 2025-06-03
omero.testlib: rename test_dir utility as create_managed_dir

sbesson:
See the failures of  https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/407/ for more context

With the release of pytest 8.4.0, tests that return a value other than None are now failing - see
https://github.com/pytest-dev/pytest/releases/tag/8.4.0. This behavior change affects all the subclasses of AbstractRepoTest. This commit renames the utility method as create_test_dir(). If this was flagged to be a backwards-compatibility concern during code review, we could keep `test_dir()` and simply `pass` or return None. Nevertheless, downstream tests consuming this method would still need to adjust their logic.

### DIFF
--- a/src/omero/testlib/__init__.py
+++ b/src/omero/testlib/__init__.py
@@ -1240,9 +1240,9 @@ PFS = (
 class AbstractRepoTest(ITest):
 
     def setup_method(self, method):
-        self.unique_dir = self.test_dir()
+        self.unique_dir = self.create_test_dir()
 
-    def test_dir(self, client=None):
+    def create_test_dir(self, client=None):
         if client is None:
             client = self.client
         mrepo = self.get_managed_repo(client=client)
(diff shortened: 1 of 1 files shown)


sbesson:
> So the test_dir() won't now be picked-up and run as it's own test, because create_managed_dir() doesn't start with test...?

Exactly. As per pytest [discovery rules](https://docs.pytest.org/en/stable/explanation/goodpractices.html#test-discovery), methods prefixed with `test` are all considrred as tests to execute.

> Does this PR need any changes in other tests?

The OMERO.py integration tests are being adjusted as of https://github.com/ome/openmicroscopy/pull/6429. I have not audited all OME repos yet to identify whether there are other test classes that would need to be updated.

sbesson:
@jburel thoughts about including this in `5.20.1` as well ?

jburel (approved):
Tests are back to green
So it makes sense to also include this PR in the upcoming release

Issue #463, opened 2025-06-30
getObjects("MapAnnotation) returns objects of different types


dominikl:
See https://forum.image.sc/t/omero-getobjects-mapannotation-returns-objects-of-different-types/114050

xfolka:
Hello!
I am the person who reported it on image.sc. Please, let me know if there is something i can do to help.

PR #464, opened 2025-06-30
Overwrite _getQueryString

dominikl:
Trying to fix https://github.com/ome/omero-py/issues/463 .

According to @will-moore the wrapper should overwrite _getQueryString. Indeed that's missing. But still also with the change, `getObjects("MapAnnotation")` returns MapAnnotationWrappers with all kind of annotations. Something else is still missing.

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -5910,6 +5910,23 @@ class MapAnnotationWrapper (AnnotationWrapper):
     """
     OMERO_TYPE = MapAnnotationI
 
+    @classmethod
+    def _getQueryString(cls, opts=None):
+        """
+        Used for building queries in generic methods such as
+        getObjects("MapAnnotation").
+        Returns a tuple of (query, clauses, params).
+
+        :param opts:        Dictionary of optional parameters.
+                            NB: No options supported for this class.
+        :return:            Tuple of string, list, ParametersI
+        """
+
+        query = ("select obj from MapAnnotation obj "
+                 "join fetch obj.details.owner as owner "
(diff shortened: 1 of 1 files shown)


will-moore:
I added a test at https://github.com/ome/openmicroscopy/pull/6435 and the fix in this PR was sufficient to get the test passing.
Let's exclude this PR to check the test fails, and then check that this PR fixes the test.

dominikl:
It still fails: https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/lastBuild/testReport/junit/OmeroPy.test.integration.gatewaytest.test_get_objects/TestFindObject/testFindAnnotation/
```
...
AssertionError: assert False
 +  where False = isinstance(object #0 (::omero::model::FileAnnotation)\n...
```

That's exactly what I got when testing it manually. It still returns all sort of annotations, not just map annotations. But I don't know why / what else is missing...

will-moore:
With this PR excluded, new test is failing as expected: https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/449/testReport/junit/OmeroPy.test.integration.gatewaytest.test_get_objects/TestFindObject/testFindAnnotation/

I'll remove the exclusion again now...

dominikl:
👍 Thanks @will-moore !

sbesson:
From my side, I think the most outstanding thing would be to agree on the strategy & timeline for resolving https://github.com/ome/omero-py/issues/465.

It would be probably a good idea to organise a review of all open PRs, close the ones that are either stale or will not be considered for inclusion and distribute the review of the candidate ones.

Issue #469, opened 2025-07-31
Move all modules under `omero`


multimeric:
Currently the `omero-py` package dumps a lot of files into the top level of `site-packages`:
<details>

```bash
ls -1 venv/lib/python3.12/site-packages/omero*.py
venv/lib/python3.12/site-packages/omero_API_ice.py
venv/lib/python3.12/site-packages/omero_Collections_ice.py
venv/lib/python3.12/site-packages/omero_Constants_ice.py
venv/lib/python3.12/site-packages/omero_FS_ice.py
venv/lib/python3.12/site-packages/omero_Internal_ice.py
venv/lib/python3.12/site-packages/omero_ModelF_ice.py
venv/lib/python3.12/site-packages/omero_ROMIO_ice.py
venv/lib/python3.12/site-packages/omero_RTypes_ice.py
venv/lib/python3.12/site-packages/omero_Repositories_ice.py
venv/lib/python3.12/site-packages/omero_Scripts_ice.py
venv/lib/python3.12/site-packages/omero_ServerErrors_ice.py
venv/lib/python3.12/site-packages/omero_ServicesF_ice.py
venv/lib/python3.12/site-packages/omero_SharedResources_ice.py
venv/lib/python3.12/site-packages/omero_SystemF_ice.py
venv/lib/python3.12/site-packages/omero_System_ice.py
venv/lib/python3.12/site-packages/omero_Tables_ice.py
venv/lib/python3.12/site-packages/omero_api_Exporter_ice.py
venv/lib/python3.12/site-packages/omero_api_IAdmin_ice.py
venv/lib/python3.12/site-packages/omero_api_IConfig_ice.py
venv/lib/python3.12/site-packages/omero_api_IContainer_ice.py
venv/lib/python3.12/site-packages/omero_api_ILdap_ice.py
venv/lib/python3.12/site-packages/omero_api_IMetadata_ice.py
venv/lib/python3.12/site-packages/omero_api_IPixels_ice.py
venv/lib/python3.12/site-packages/omero_api_IProjection_ice.py
venv/lib/python3.12/site-packages/omero_api_IQuery_ice.py
venv/lib/python3.12/site-packages/omero_api_IRenderingSettings_ice.py
venv/lib/python3.12/site-packages/omero_api_IRepositoryInfo_ice.py
venv/lib/python3.12/site-packages/omero_api_IRoi_ice.py
venv/lib/python3.12/site-packages/omero_api_IScript_ice.py
[...]

PR #470, opened 2025-07-31
Faux import

multimeric:
Another, simpler attempt at #448.

The logic behind this is that `omero.model` imports classes from `omero_model_XXX` but type checkers can't understand this because it's done dynamically. If we put the imports into a `if TYPE_CHECKING` guard, it won't run at runtime but type checkers will at least understand the link.

This approach still lacks actual type annotations, but to solve this we could need to address #469 and/or https://github.com/zeroc-ice/ice/issues/4264.

Mypy probably still won't like this, because there are a lot of other type issues including the fact it can't statically analyse the parent class relationships which requires https://github.com/zeroc-ice/ice/issues/4265. Still, this should be an improvement because you can list and navigate to the class definitions in VS Code:
Screenshot 2025-07-31 at 4 39 15 PM

The pylance server also seems to understand which methods the class has:
Screenshot 2025-07-31 at 4 42 17 PM

But it doesn't understand instance methods because of https://github.com/zeroc-ice/ice/issues/4265.
Screenshot 2025-07-31 at 4 42 37 PM


multimeric:
Actually you can fix the above error by adding a tiny `Ice` stub:

```bash
$ cat typings/Ice/Ice.pyi
```
```python
from typing import Any

def openModule(module: str) -> Any: ...
```

Screenshot 2025-07-31 at 5 35 53 PM

In theory this should be fixed upstream in `Ice`, but we're using 3.6 so any upstream changes won't reach us anyway.

multimeric:
That's an interesting question, and probably the best solution here. I'll open another issue.

sbesson:
@pepone from a maintenance perspective, we are aware that Ice will need to be upgraded at some stage. From previous experiences, we are arguably talking of the most challenging OMERO dependency. The complexity here is that the impact of the library upgrade goes way beyond the scope of OMERO.py as similar changes will need to be applied to OMERO server as well as every OMERO client. Practically, we are talking of a backwards-incompatible change affecting the whole OMERO ecosystem associated with a significant amount of scoping, work and coordination both on the development & testing as well as the deployment side.  At present, there is no planning of such effort by the OME & Glencoe Software teams.

Specifically for OMERO.py, the Python 3.12+ requirement for Ice 3.8 is another concern as it is more restrictive than the minimal requirement for this library which is currently Python 3.9+, probably moving to 3.10+ later in the year as Python 3.9 goes officially EOL.

joshmoore:
Since omero-py wouldn't be installed with Ice>3.6 for the moment, would it be safe to ship `typings/Ice/Ice.pyi` here?

multimeric:
I suppose an alternative to this PR is for users to just import the models directly from their "true" location. ie instead of:
```python
from omero import gateway, model
gateway.DatasetWrapper(conn, model.DatasetI())
```

You can do the following, which type checkers can better understand:
```python
from omero import gateway
from omero_model_DatasetI import DatasetI

gateway.DatasetWrapper(conn, DatasetI())
```

PR #390, opened 2024-01-18
Removal of python-future compatibility code

sbesson:
Fixes #389 

Primarily motivated by the support of Python 3.12 in OMERO and given the current maintenance state of the [python-future](https://python-future.org/) project, this reviews the OMERO.py source code to update all paths with either Python 2 or Python 2/3 logic to pure Python 3+.

- all imports from the `future` dependency have been reviewed and cleaned up. The commits should contain some details but capturing a few specific notes on updates that might require particular attention during the review
  * `isbytes` and `bytes_to_native_str` have been replaced by `isinstance(x, bytes)` and `x.decode('utf-8')`
  * `old_div` has been replaced by `/` for path concatenation and either `/` or `//` for mathematical divisions depending on the use case
  * `long` has been removed as it is no longer a basic type in Python 3
  * `basestring` has been replaced by `str` and/or `bytes` depending on the use case
- all imports `builtins` have been cleaned up as they should be unnecessary
- all Python 2 codepaths using the value of `sys.version_info` have been cleaned up

This branch has been minimally tested in an Ubuntu 22.04 deployment (Python 3.10) with OMERO.server and OMERO.web. There are a few initial questions:
- should the `future` dependency be fully removed from `setup.py` or kept for compatibility with the downstream projects?
- considering team testing, several downstream Python projects including OMERO.web, openmicroscopy, various OMERO CLI plugins might need a similar handling. How many do we want to be patched in a similar fashion before scheduling a round of functional testing?

Initially opening as a draft PR to allow and gather feedback. I propose to get in the nightly CI builds over the week-end to get extra coverage from the integration tests and fix additional outstanding issues and we can discuss the progress next week.

PR #478, opened 2025-09-22
Remove Python 2 compat from gateway utilities and scripts tests

chris-allan:
Removes the Python 2 compatibility try block from the gateway utilities and scripts tests. There hasn't been Python 2 support for years.

Other modernisation was done in #390.


jburel:
https://github.com/ome/omero-py/blob/master/test/unit/scriptstest/test_parse.py#L25 should be fixed as well

PR #453, opened 2025-03-13
getObject() and getExternalInfo() load externalInfo

will-moore:
See https://github.com/ome/omero-web/pull/613.

This tweaks the base `BlitzObjectWrapper._getQueryString()` to include loading `externalInfo`.

It also adds `BlitzObjectWrapper.getExternalInfo()` to return the `externalInfo` and load it if necessary, e.g. for cases where `self._obj` is NOT the object loaded by `getObject()`.

E.g. for image:

```
def __loadedHotSwap__(self):
        ctx = self._conn.SERVICE_OPTS.copy()
        ctx.setOmeroGroup(self.getDetails().group.id.val)
        self._obj = self._conn.getContainerService().getImages(
            self.OMERO_CLASS, (self._oid,), None, ctx)[0]
```

To test:

Set some values for ExternalInfo:
Using this branch:

```
$ omero obj ext-info-set Image:12 myLsid myEntityType
```

Since the JSON API uses `_getQueryString()` to build queries, and omero-marshal already handles externalInfo, you can check the details at E.g. `/api/v0/m/images/ID/`: (NB: deployed on idr-testing, e.g. `/api/v0/m/images/15159664/`).

```
"externalInfo": {
    "@type": "TBD#ExternalInfo",
    "@id": 8,
    "omero:details": {},
    "EntityId": 3,
    "EntityType": "com.glencoesoftware.ngff:multiscales",
    "Lsid": "test_dataset_extinfo"
}
```

PR #482, opened 2025-09-30
omero.gateway: fix DetailsWrapper.getExternalInfo regression

### LINKED ISSUE #481: Python gateway: ExternalInfo getter regression
In OMERO.py 5.21.0, [#453](https://github.com/ome/omero-py/pull/453) introduced a new `getExternalInfo` API allowing to lazy-load `ExternalInfo` objects available from the details of various OMERO model objects.

The addition of this API caused some regression that can be demonstrated in the following example:

```python
import omero.gateway
conn = omero.gateway.BlitzGateway(host=host,username=username,passwd=password)
try:
    conn.connect()
    image = conn.getObject("Image", iamge_idid)
    ext_info = image.getDetails().getExternalInfo()
finally:
    conn.close()
```

With OMERO.py 5.20.0 executing the above against a running server and valid image ID works but with OMERO.py 5.21.0 or 5.21.1 it will fail with the following stack trace

```
  File "/Users/sbesson/Documents/GitHub/omero-py/test.py", line 7, in <module>
    ext_info = image.getDetails().getExternalInfo()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sbesson/Documents/GitHub/omero-py/venv/lib/python3.12/site-packages/omero/gateway/__init__.py", line 1355, in getExternalInfo
    extinfo = self.getDetails()._externalInfo
              ^^^^^^^^^^^^^^^^^
  File "/Users/sbesson/Documents/GitHub/omero-py/venv/lib/python3.12/site-packages/omero/gateway/__init__.py", line 472, in getDetails
    if self._obj.loaded:
       ^^^^^^^^^^^^^^^^
  File "/Users/sbesson/Documents/GitHub/omero-py/venv/lib/python3.12/site-packages/omero_model_DetailsI.py", line 106, in __getattr__
    raise AttributeError(attr)
AttributeError: loaded
```

The underlying issue is that `getExternalInfo` API was defined at the `BlitzObjectWrapper` level and is inherited by all instance wrappers. This includes `DetailsWrapper` and overrides the previous getter which was either returning `None` or a `BlitzObjectWrapper` object depending on the use case.

One way to fix this is to override `getExternalInfo` again in `DetailsWrapper` restoring the previous behavior. @will-moore do you know if there is a better option to limit the `getExternalInfo` API to the relevant wrappers?

sbesson:
Fixes #481 

6e521bba8be7e8f2e22b14fcf309e1d703b5ecdb adds a simple unit test capturing the regression. The test should pass with OMERO.py 5.20.0 but fail with OMERO.py 5.21.0 or 5.21.1.

1fd4c4d2fe9af68524d529fb0ec9bf40a92560fd fixes the regression by overriding `BlitzObjectWrapper.getExternalInfo()` for `DetailsWrapper`

69841a510b3f09b50c1a3966c5b99a84e965637b adds some coverage for the API introduced in #453 

The tests have been initially limited to the scope of `ImageWrapper` but happy to extend them to cover other wrappers if there is an authoritative list of the types we'd like to support.

### DIFF
--- a/test/unit/test_gateway.py
+++ b/test/unit/test_gateway.py
@@ -357,6 +357,7 @@ class TestBlitzGatewayImageWrapper(object):
         assert wrapper.getDetails().getExternalInfo() is None
         assert wrapper._obj.getDetails().getExternalInfo() is None
         assert wrapper.getDetails()._obj.getExternalInfo() is None
+        assert wrapper.getExternalInfo() is None
 
         external_info = ExternalInfoI()
         external_info.id = rlong(1)
@@ -368,3 +369,6 @@ class TestBlitzGatewayImageWrapper(object):
         assert wrapper.getDetails().getExternalInfo()._obj == external_info
         assert wrapper._obj.getDetails().getExternalInfo() == external_info
         assert wrapper.getDetails()._obj.getExternalInfo() == external_info
+        # BlitzObjectWrapper.getExternalInfo() returns omero.model.ExternalInfo
+        assert isinstance(wrapper.getExternalInfo(), ExternalInfoI)
+        assert wrapper.getExternalInfo() == external_info


will-moore (approved):
Looks good, thx.

PR #489, opened 2025-11-24
Add more opts support for getObjects('Annotation')

### LINKED ISSUE #433: BlitzGateway: Support linking annotations to other annotations
A user came across this issue when trying to programatically create TagAnnotations linked under a parent TagSet.

Currently the [linkAnnotation](https://github.com/ome/omero-py/blob/master/src/omero/gateway/__init__.py#L1049-L1059) function on the Blitz wrapper classes uses the `OMERO_CLASS` class variable to construct the link object name. Since `Annotation` wrapper subclasses do not define this variable the function errors out searching for `NoneAnnotationLink`.

It looks like setting `OMERO_CLASS = "Annotation"` would fix this particular scenario (we need an `AnnotationAnnotationLink`), but it's not clear to me if this would cause problems with other functionality in the wrapper.

will-moore:
# conn.getObjects("Annotation")

This adds support for various options for `conn.getObjects("Annotation", opts)`, in preparation for JSON api:
The `OMERO_CLASS` is now set on `AnnotationWrapper` objects.
Also Fixes #433.

This is used by https://github.com/ome/omero-web/pull/682

Tests added in https://github.com/ome/openmicroscopy/pull/6458

Supported opts - All are optional (but parent_ids needs parent_type)

 - `"parent_type": "dataset"`
 - `"parent_ids": [1, 2]`
 - `"ns": "my.namespace"`

E.g. `/api/tagannotations/?project=1&project=2` would be backed by:

```
conn.getObjects("TagAnnotation", opts={
  "parent_type": "project",
  "parent_ids": [1,2]
})
```

# %sAnnotationLink

This also improves the formulation of `%sAnnotationLink` (E.g. `ImageAnnotationLink`, `ExperimenterGroupAnnotationLink`, `AnnotationAnnotationLink`) in multiple places:

 - `conn.getObjectsByMapAnnotations("Dataset", key="foo")`
 - `obj._loadAnnotationLinks() / _getAnnotationLinks()`
   - `unlinkAnnotations()`
   - `removeAnnotations()`
   - `getAnnotation()`
   - `listAnnotations()`
 - `obj._linkAnnotation() / _linkObject` unused?
 - `conn.getAnnotationLinks()`
 - `conn.countAnnotations()`
 - `conn.listOrphanedAnnotations()`
 - `AnnotationWrapper._getQueryString()` 

We can now do `getObject()` and `listAnnotations()` etc with many more object types:

E.g:

```
tag = omero.gateway.TagAnnotationWrapper(conn)
tag.setValue("Test Tag on Group")

group = conn.getGroupFromContext()
group.linkAnnotation(tag)

group.listAnnotations()
```

# Known issue: Loading Parents

For webclient/api/annotations/ we get the annotations along with links to the parents and the parent objects too. 
E.g. 

```
  "link": {
    "id": 39752336,
    "owner": {
      "id": 2
    },
    "parent": {
      "id": 3414011,
      "class": "ImageI",
      "name": "10percent-Wt1-GFP-spheroid-MV.czi [0]"
    },
    "date": "2017-11-27T14:13:19+00:00",
    "permissions": {
      "canDelete": false,
      "canAnnotate": false,
      "canLink": false,
      "canEdit": false
    }
  }
```

This is useful to have permissions, timestamp, owner etc on the link but it is particularly essential when we are loading annotations on multiple objects. E.g. `/api/annotations/?type=tag&project=1&project=2` would load annotations for 2 projects, BUT we without links, we don't know which Annotations are linked to which Project.

But, it doesn't appear possible to achieve this with `omero_marshal`, since for Annotations, there is no handling of links to Parent objects. Annotations are "Annotatable", so we check for annotations ON the object https://github.com/ome/omero-marshal/blob/f8ff1c2e439f0599d96423b7f64898864548ba9f/omero_marshal/encode/encoders/annotation.py#L68
but not whether the Object (annotation) is annotating a Parent.
In fact, I don't see that it's possible for e.g. an Annotation to have links loaded (in the same way that a Project or Dataset can have `obj.copyAnnotationLinks()`, since the possible links for an Annotation object are stored in a different Table for each parent, e.g. `ProjectAnnotationLinks, DatasetAnnotationLinks` etc.

### DIFF
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -5162,10 +5162,42 @@ class AnnotationWrapper (BlitzObjectWrapper):
                             NB: No options supported for this class.
         :return:            Tuple of string, list, ParametersI
         """
+
+        query, clauses, params = super(
+            AnnotationWrapper, cls)._getQueryString(opts)
+        if opts is None:
+            opts = {}
+
         query = ("select obj from Annotation obj "
                  "join fetch obj.details.owner as owner "
                  "join fetch obj.details.creationEvent")
-        return query, [], omero.sys.ParametersI()
+
+        ann_type = opts.get('ann_type', None)
+        if ann_type == "tag":
+            clauses.append("obj.class=TagAnnotation")
+        elif ann_type == "file":
+            clauses.append("obj.class=FileAnnotation")
+        elif ann_type == "comment":
+            clauses.append("obj.class=CommentAnnotation")
+        elif ann_type == "rating":
+            clauses.append("obj.class=LongAnnotation")
+            clauses.append("obj.ns='openmicroscopy.org/omero/insight/rating'")
+        elif ann_type == "map":
+            clauses.append("obj.class=MapAnnotation")
+
+        for obj_type in ['Project', 'Dataset', 'Image', 'Screen',
+                         'Plate', 'Well', 'Roi']:
+            plural = obj_type.lower() + "s"
+            if plural in opts:
+                ids = opts[plural]
+                if isinstance(ids, list) and len(ids) > 0:
+                    clauses.append(
+                        "exists (from %sAnnotationLink as link "
+                        "where link.child.id = obj.id "
+                        "and link.parent.id in (:%s_ids))" % (obj_type, plural))
+                    params.add("%s_ids" % plural, rlist([rlong(i) for i in ids]))
+        
+        return (query, clauses, params)
 
     @classmethod
     def _register(cls, regklass):
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -5164,7 +5164,9 @@ class AnnotationWrapper (BlitzObjectWrapper):
         Returns a tuple of (query, clauses, params).
 
         :param opts:        Dictionary of optional parameters.
-                            NB: No options supported for this class.
+                            ann_type: "tag", "file", "comment", "rating", "map"
+                            parent_type: (optional) "project", "dataset", "image" etc
+                            parent_ids: (optional) list of IDs for the parent type
         :return:            Tuple of string, list, ParametersI
         """
 
@@ -5190,18 +5192,17 @@ class AnnotationWrapper (BlitzObjectWrapper):
         elif ann_type == "map":
             clauses.append("obj.class=MapAnnotation")
 
-        for obj_type in ['Project', 'Dataset', 'Image', 'Screen',
-                         'Plate', 'Well', 'Roi']:
-            plural = obj_type.lower() + "s"
-            if plural in opts:
-                ids = opts[plural]
-                if isinstance(ids, list) and len(ids) > 0:
-                    clauses.append(
-                        "exists (from %sAnnotationLink as link "
-                        "where link.child.id = obj.id "
-                        "and link.parent.id in (:%s_ids))" % (obj_type, plural))
-                    params.add("%s_ids" % plural, rlist([rlong(i) for i in ids]))
-        
+        if 'parent_type' in opts:
+            obj_type = opts['parent_type'].title().replace("Plateacquisition", "PlateAcquisition")
+            ids_clause = ""
+            if 'parent_ids' in opts:
+                ids_clause = f"and link.parent.id in (:parent_ids)"
+                params.add("parent_ids", rlist([rlong(i) for i in opts['parent_ids']]))
+
+            clause = f"""exists (from {obj_type}AnnotationLink as link
+                        where link.child.id = obj.id {ids_clause})"""
+            clauses.append(clause)
+
         return (query, clauses, params)
 
     @classmethod
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -10756,6 +10756,10 @@ class _LightSourceWrapper (BlitzObjectWrapper):
               '#type;lightSourceType',
               'version')
 
+    @classmethod
+    def ann_link_name(klass):
+        return "LightSourceAnnotationLink"
+
     def getLightSourceType(self):
         """
         Gets the Light Source type for this light source (enum value)
@@ -11053,7 +11057,6 @@ def refreshWrappers():
                            "filter": FilterWrapper,
                            "instrument": InstrumentWrapper,
                            "lightpath": LightPathWrapper,
-                           "lightsource": LightSourceWrapper,
                            "objective": ObjectiveWrapper,
                            "planeinfo": PlaneInfoWrapper,
                            # allows for getObjects("Annotation", ids)
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -5165,7 +5165,7 @@ class AnnotationWrapper (BlitzObjectWrapper):
         Returns a tuple of (query, clauses, params).
 
         :param opts:        Dictionary of optional parameters.
-                            parent_type: (optional) "project", "dataset", "image" etc
+                            parent_type: (optional) "Project", "Dataset", "Image" etc
                             parent_ids: (optional) list of IDs for the parent type
                             ns: (optional) namespace string to filter by
         :return:            Tuple of string, list, ParametersI
@@ -5185,8 +5185,15 @@ class AnnotationWrapper (BlitzObjectWrapper):
                  "join fetch obj.details.owner as owner "
                  "join fetch obj.details.creationEvent")
 
+        # We want to make parent_type case-insensitive...
         if 'parent_type' in opts:
-            obj_type = opts['parent_type'].title().replace("Plateacquisition", "PlateAcquisition")
+            # Title case works for most objects...
+            obj_type = opts['parent_type'].title()
+            # Handle special cases for certain annotatable object types
+            for otype in ["ExperimenterGroup", "LightPath", "LightSource",
+                          "OriginalFile", "PlaneInfo", "PlateAcquisition"]:
+                if obj_type == otype.title():
+                    obj_type = otype
             ids_clause = ""
             if 'parent_ids' in opts:
                 ids_clause = f"and link.parent.id in (:parent_ids)"
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -169,23 +169,7 @@ def getAnnotationLinkTableName(objecttype):
     Get the name of the AnnotationLink table
     for the given objecttype
     """
-    objecttype = objecttype.lower()
-
-    if objecttype == "project":
-        return "ProjectAnnotationLink"
-    if objecttype == "dataset":
-        return"DatasetAnnotationLink"
-    if objecttype == "image":
-        return"ImageAnnotationLink"
-    if objecttype == "screen":
-        return "ScreenAnnotationLink"
-    if objecttype == "plate":
-        return "PlateAnnotationLink"
-    if objecttype == "plateacquisition":
-        return "PlateAcquisitionAnnotationLink"
-    if objecttype == "well":
-        return "WellAnnotationLink"
-    return None
+    return ann_link_name(objecttype)
 
 
 def getPixelsQuery(imageName):
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -5613,7 +5613,7 @@ from omero_model_TagAnnotationI import TagAnnotationI
 
 class TagAnnotationWrapper (AnnotationWrapper):
     """
-    omero_model_TagAnnotationI class wrapper extends AnnotationWrapper.
+    omero_model_BooleanAnnotationI class wrapper extends AnnotationWrapper.
     """
 
     OMERO_TYPE = TagAnnotationI
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -1062,7 +1062,11 @@ class BlitzObjectWrapper (object):
         :param ann:     The annotation object
         :type ann:      :class:`AnnotationWrapper`
         """
-        return self._linkObject(ann, "%sAnnotationLinkI" % self.OMERO_CLASS)
+        class_string = self.OMERO_CLASS
+        # e.g. "TagAnnotation" -> "AnnotationAnnotationLinkI"
+        if class_string.endswith("Annotation"):
+            class_string = "Annotation"
+        return self._linkObject(ann, "%sAnnotationLinkI" % class_string)
 
     def linkAnnotation(self, ann, sameOwner=False):
         """
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -5164,9 +5164,10 @@ class AnnotationWrapper (BlitzObjectWrapper):
         Returns a tuple of (query, clauses, params).
 
         :param opts:        Dictionary of optional parameters.
-                            ann_type: "tag", "file", "comment", "rating", "map"
+                            ann_type: (optional) "tag", "file", "comment", "rating", "map"
                             parent_type: (optional) "project", "dataset", "image" etc
                             parent_ids: (optional) list of IDs for the parent type
+                            ns: (optional) namespace string to filter by
         :return:            Tuple of string, list, ParametersI
         """
 
@@ -5202,6 +5203,10 @@ class AnnotationWrapper (BlitzObjectWrapper):
             clause = f"""exists (from {obj_type}AnnotationLink as link
                         where link.child.id = obj.id {ids_clause})"""
             clauses.append(clause)
+
+        if 'ns' in opts:
+            clauses.append("obj.ns=:ns")
+            params.add("ns", rstring(opts['ns']))
 
         return (query, clauses, params)
 
--- a/src/omero/gateway/__init__.py
+++ b/src/omero/gateway/__init__.py
@@ -910,13 +910,13 @@ class BlitzObjectWrapper (object):
         ctx = self._conn.SERVICE_OPTS.copy()
         ctx.setOmeroGroup(self.details.group.id.val)
         if not self._obj.isAnnotationLinksLoaded():
-            query = ("select l from %sAnnotationLink as l join "
+            query = ("select l from %s as l join "
                      "fetch l.details.owner join "
                      "fetch l.details.creationEvent "
                      "join fetch l.child as a join fetch a.details.owner "
                      "left outer join fetch a.file "
                      "join fetch a.details.creationEvent where l.parent.id=%i"
-                     % (self.OMERO_CLASS, self._oid))
+                     % (self.ann_link_name(), self._oid))
             links = self._conn.getQueryService().findAllByQuery(
                 query, None, ctx)
             self._obj._annotationLinksLoaded = True
@@ -1066,11 +1066,8 @@ class BlitzObjectWrapper (object):
         :param ann:     The annotation object
         :type ann:      :class:`AnnotationWrapper`
         """
-        class_string = self.OMERO_CLASS
-        # e.g. "TagAnnotation" -> "AnnotationAnnotationLinkI"
-        if class_string.endswith("Annotation"):
-            class_string = "Annotation"
-        return self._linkObject(ann, "%sAnnotationLinkI" % class_string)
+        link_name = self.ann_link_name()
+        return self._linkObject(ann, "%sI" % link_name)
 
     def linkAnnotation(self, ann, sameOwner=False):
         """
@@ -3556,18 +3553,15 @@ class _BlitzGateway (object):
                        "Must be one of: %s" % (parent_type, wrapper_types))
             raise AttributeError(err_msg)
         wrapper = KNOWN_WRAPPERS.get(parent_type.lower(), None)
-        class_string = wrapper().OMERO_CLASS
-        # E.g. TagAnnotation -> Annotation for AnnotationAnnotationLink
-        if class_string.endswith("Annotation"):
-            class_string = "Annotation"
+        link_name = wrapper.ann_link_name()
 
-        query = ("select annLink from %sAnnotationLink as annLink "
+        query = ("select annLink from %s as annLink "
                  "join fetch annLink.details.owner as owner "
                  "join fetch annLink.details.creationEvent "
                  "join fetch annLink.child as ann "
                  "join fetch ann.details.owner "
                  "join fetch ann.details.creationEvent "
-                 "join fetch annLink.parent as parent" % class_string)
+                 "join fetch annLink.parent as parent" % link_name)
 
         q = self.getQueryService()
         if params is None:
@@ -3619,6 +3613,8 @@ class _BlitzGateway (object):
         if obj_type is None or not obj_ids:
             return counts
 
+        wrapper = KNOWN_WRAPPERS.get(obj_type.lower(), None)
+
         ctx = self.SERVICE_OPTS.copy()
         ctx.setOmeroGroup(-1)
 
@@ -3637,12 +3633,12 @@ class _BlitzGateway (object):
                sum( case when an.class != LongAnnotation
                         then 1 else 0 end ), type(an.class)
                from Annotation an where an.id in
-                    (select distinct(ann.id) from %sAnnotationLink ial
+                    (select distinct(ann.id) from %s ial
                         join ial.child as ann
                         join ial.parent as i
                 where i.id in (:ids))
                     group by an.class
-            """ % obj_type
+            """ % wrapper.ann_link_name()
 
         queryResult = self.getQueryService().projection(q, params, ctx)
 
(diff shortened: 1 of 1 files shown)


will-moore:
Tests added in https://github.com/ome/openmicroscopy/pull/6458

will-moore:
Comment on what changed after first commit....

The problem with the `getObjects()` is that we can't easily handle `OR` logic. E.g. "get Annotations with links to Project:1 OR Dataset:2". Currently, for each `obj_type` we add a clause, so we get `AND` logic:

```
                    clauses.append(
                        "exists (from %sAnnotationLink as link "
                        "where link.child.id = obj.id "
                        "and link.parent.id in (:%s_ids))" % (obj_type, plural))
                    params.add("%s_ids" % plural, rlist([rlong(i) for i in ids]))
```
Therefore, 
`conn.getObjects("Annotation", opts={"ann_type": "tag", "projects": [1,2], "datasets": [3,4]})` won't give you annotations on all objects, as you currently get with e.g. https://idr.openmicroscopy.org/webclient/api/annotations/?type=map&screen=3&image=3414011

Do we ever need to load annotations for more that 1 type of Object? E.g. webclient almost never allows you to select multiple different Object types, (e.g. Dataset and Image, except maybe in the search results)

The webclient/api endpoint does a series of queries `select oal from %sAnnotationLink as oal` for each obj_type requested. Each Database query only supports 1 object type. So it's probably OK for the `getObjects("Annotation")` to also support just ONE parent type. Same behaviour in E.g. 

`conn.getAnnotationLinks(...parent_type, parent_ids=None)`

So, let's instead go for:

```
conn.getObjects("Annotation", opts={
  "ann_type": "tag",
  "parent_type": "project",
  "parent_ids": [1,2]
})
```

knabar:
> Do we ever need to load annotations for more that 1 type of Object? E.g. webclient almost never allows you to select multiple different Object types, (e.g. Dataset and Image, except maybe in the search results)
>
> The webclient/api endpoint does a series of queries `select oal from %sAnnotationLink as oal` for each obj_type requested. Each Database query only supports 1 object type. So it's probably OK for the `getObjects("Annotation")` to also support just ONE parent type. 

It makes sense to me to limit the parent type to one per request, if a client requires annotations on different parent types, the client can make multiple requests.

I think it could even be justified to limit requests to a single parent object if that simplifies the requests and responses significantly.

src/omero/gateway/__init__.py
```diff
@@ -5164,13 +5164,54 @@ def _getQueryString(cls, opts=None):
...
-        return query, [], omero.sys.ParametersI()
+
+        ann_type = opts.get('ann_type', None)
+        if ann_type == "tag":
+            clauses.append("obj.class=TagAnnotation")
+        elif ann_type == "file":
+            clauses.append("obj.class=FileAnnotation")
+        elif ann_type == "comment":
+            clauses.append("obj.class=CommentAnnotation")
+        elif ann_type == "rating":
```
knabar: Similar to parent types, it might be best to stick with the built-in annotation types here. There is only one other reference to ratings in all of OMERO.py, which arguably should not be there either. This would also remove the special treatment of the namespace option in line 5210.

will-moore: Done in 367e22e

will-moore:
@knabar Anything else to address here?

> I think it could even be justified to limit requests to a single parent object if that simplifies the requests and responses significantly.

There is no additional logic associated with supporting multiple parent objects (IDs) of the same type, and it could be useful so I'd like to leave this in.

src/omero/gateway/__init__.py
```diff
@@ -5164,13 +5164,55 @@ def _getQueryString(cls, opts=None):
...
+        elif ann_type == "file":
+            clauses.append("obj.class=FileAnnotation")
+        elif ann_type == "comment":
+            clauses.append("obj.class=CommentAnnotation")
+        elif ann_type == "long":
+            clauses.append("obj.class=LongAnnotation")
+        elif ann_type == "map":
+            clauses.append("obj.class=MapAnnotation")
+        elif ann_type is not None:
+            msg = ("ann_type '%s' not recognised. Must be one of 'tag', 'file', "
```
sbesson: This should be implemented for all supported annotations (BooleanAnnotation, TermAnnotation, XmlAnnotation...)

src/omero/gateway/__init__.py
```diff
@@ -5164,13 +5164,55 @@ def _getQueryString(cls, opts=None):
         Returns a tuple of (query, clauses, params).
 
         :param opts:        Dictionary of optional parameters.
-                            NB: No options supported for this class.
+                            ann_type: (optional) "tag", "file", "comment", "long", "map"
```
sbesson: Arguably this could use the full annotation type `TagAnnotation`, `FileAnnotation` and save the mapping below.

src/omero/gateway/__init__.py
```diff
@@ -5164,13 +5164,55 @@ def _getQueryString(cls, opts=None):
...
+                            parent_type: (optional) "project", "dataset", "image" etc
```
sbesson: Is there a list of all objects that can be annotated?

src/omero/gateway/__init__.py
```diff
@@ -5164,13 +5164,55 @@ def _getQueryString(cls, opts=None):
...
+            clauses.append("obj.class=LongAnnotation")
+        elif ann_type == "map":
+            clauses.append("obj.class=MapAnnotation")
+        elif ann_type is not None:
+            msg = ("ann_type '%s' not recognised. Must be one of 'tag', 'file', "
+                   "'comment', 'long', 'map'" % ann_type)
+            raise AttributeError(msg)
+
+        if 'parent_type' in opts:
+            obj_type = opts['parent_type'].title().replace("Plateacquisition", "PlateAcquisition")
```
sbesson: Unfortunately, `PlateAcquisition` is not the only object that will need this special handling to convert from its lowercase version. Other examples include all annotation types (which can be annotated themselves), some instrument objects (LightPath, LightSource...) as well as OriginalFile.

sbesson (changes requested):
A few initial comments. I think the extensibility to support the filtering by annotation type, parent and namespace is useful but we will need additional changes to handle more generically all relevant objects.

src/omero/gateway/__init__.py
will-moore: I don't actually know where to find such a list or if it's possible to create one?

will-moore: https://omero.readthedocs.io/en/stable/developers/Model/EveryObject.html#annotation seems like a good start...

src/omero/gateway/__init__.py
will-moore: Using this list (from https://omero.readthedocs.io/en/stable/developers/Model/EveryObject.html#annotation)

```
AnnotationAnnotationLink
ChannelAnnotationLink
DatasetAnnotationLink
DetectorAnnotationLink
DichroicAnnotationLink
ExperimenterAnnotationLink
ExperimenterGroupAnnotationLink
FilesetAnnotationLink
FilterAnnotationLink
FolderAnnotationLink
ImageAnnotationLink
InstrumentAnnotationLink
LightPathAnnotationLink
LightSourceAnnotationLink
NamespaceAnnotationLink
NodeAnnotationLink
ObjectiveAnnotationLink
OriginalFileAnnotationLink
PlaneInfoAnnotationLink
PlateAcquisitionAnnotationLink
PlateAnnotationLink
ProjectAnnotationLink
ReagentAnnotationLink
RoiAnnotationLink
ScreenAnnotationLink
SessionAnnotationLink
ShapeAnnotationLink
WellAnnotationLink
```

I see only these cases:

```
ExperimenterGroup
LightPath
LightSource
OriginalFile
PlaneInfo
PlateAcquisition
```

If the `parent_type` is `annotation` then this will query for `AnnotationAnnotationLink` without any other handling needed.

src/omero/gateway/__init__.py
will-moore: Do you mean `"TagAnnotation"` (string) or `omero.model.TagAnnotation` (class)?

I feel that it's more user-friendly for the `BlitzGateway` methods to take strings rather than `omero.model` objects. E.g. `conn.getObject("Image")` rather than `conn.getObject(omero.model.ImageI)`.
So I'd like to do the mapping here. Otherwise we'd have to do the mapping in `omero-web` for https://github.com/ome/omero-web/pull/682 and probably other places too.

I have just realised that the more natural way to do this (since we only support a single ann type) is to do:

e.g. get Tags on datasets... NB: we already support `conn.getObject("TagAnnotation")`

```
conn.getObject("TagAnnotation", opts={"parent_type": "dataset", "parent_ids": [1, 2]})
```

instead of

```
conn.getObject("Annotation", opts={"ann_type": "tag", "parent_type": "dataset", "parent_ids": [1, 2]})
```

sbesson: Agreed, that's a more natural extension of the existing API and substantially simplifies the logic. 
I think it would make sense to use the same semantics for the `parent_type` e.g. when retrieving  all the tags under a tagset, one could use

```
conn.getObjects("TagAnnotation", opts={"parent_type": "TagAnnotation", "parent_ids": [1], ns: omero.constants.metadata.NSINSIGHTTAGSET})
```

This would remove the special casing discussed in https://github.com/ome/omero-py/pull/489#discussion_r3682568799

will-moore: Actually, `conn.getObjects("dataset')` or `conn.getObjects("tagannotation")` is already case-insensitive, so lets stick with that here too for consistency.

src/omero/gateway/__init__.py
```diff
@@ -5164,13 +5165,49 @@ def _getQueryString(cls, opts=None):
...
+        if 'parent_type' in opts:
+            # Title case works for most objects...
+            obj_type = opts['parent_type'].title()
+            # Handle special cases for certain annotatable object types
+            for otype in ["ExperimenterGroup", "LightPath", "LightSource",
+                          "OriginalFile", "PlaneInfo", "PlateAcquisition"]:
+                if obj_type == otype.title():
+                    obj_type = otype
+            ids_clause = ""
+            if 'parent_ids' in opts:
```
sbesson, line 5202: What is the real-world use case associated with specifying `parent_type` but no `parent_ids`? 
I think it should be easy to enforce that `parent_ids` must be supplied if `parent_type` is specified. And that can always be relaxed later on if there is a real need.

src/omero/gateway/__init__.py
```diff
@@ -5164,13 +5165,49 @@ def _getQueryString(cls, opts=None):
...
+
+        query = (f"select obj from {cls.OMERO_CLASS} obj "
+                 f"{fetch_file} "
                  "join fetch obj.details.owner as owner "
                  "join fetch obj.details.creationEvent")
-        return query, [], omero.sys.ParametersI()
+
+        # We want to make parent_type case-insensitive...
+        if 'parent_type' in opts:
+            # Title case works for most objects...
```
sbesson, line 5194: Can we move this utility to a separate utility method for transforming lower-case strings into case-sensitive object types with associated unit tests ?

I see a few places in the gateway that can use it. There might be use cases elsewhere in OMERO.py.

sbesson (changes requested):
This is getting much better in particular with the removal of the duplicated `_getQueryString` to use OMERO_CLASS (with the additional clause for fetching `OriginalFile`).

Reviewing [the list of structured annotations](https://omero.readthedocs.io/en/stable/developers/Model/StructuredAnnotations.html), the only concrete instantiation that misses a wrapper is `ListAnnotation`. Is that something we want to quickly implement as part of this PR to be feature complete?

pwalczysko:
@will-moore excluding - this PR is causing the ``test_show.py`` errors in https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/43/testReport/

sbesson:
More specifically, the addition of `OMERO_CLASS` static variable to the various annotation wrapper classes is at odds with the implementation of `
‎_BlitzGateway.getAnnotationLinks` - see https://github.com/ome/omero-py/blob/5d0974d097e25d98892b9b419528edf9256274b0/src/omero/gateway/__init__.py#L3551-L3554 which will need to be updated accordingly.

will-moore:
@sbesson Thanks for catching that. I realise (since no tests failed) that we don't have tests for this, so I added that to https://github.com/ome/openmicroscopy/pull/6458/changes/2084d18e97305bd4f6e2b4e34620eeb5fdc3c0c8
That also revealed that e.g. `ann.linkAnnotation(tag)` didn't work for linking annotations to *annotations* because `ann` is expected to have `OMERO_CLASS` so I fixed that too.

sbesson:
> That also revealed that e.g. ann.linkAnnotation(tag) didn't work for linking annotations to annotations because ann is expected to have OMERO_CLASS so I fixed that too.

:+1: does this mean this PR is incidentally fixing #433 as well?

src/omero/gateway/__init__.py
will-moore: @sbesson You may want to search for all `omero.figure.json` (ns) annotations that are on `Datasets` rather than on `Images`, or look for Tags on ROIs rather than on other objects etc. It's possible to imagine various use-cases.

will-moore:
Updated the `%sAnnotationLink` logic in these methods:

 - `conn.getObjectsByMapAnnotations("Dataset", key="foo")`
 - `obj._loadAnnotationLinks() / _getAnnotationLinks()`
   - `unlinkAnnotations()`
   - `removeAnnotations()`
   - `getAnnotation()`
   - `listAnnotations()`
 - `obj._linkAnnotation() / _linkObject` unused?
 - `conn.getAnnotationLinks()`
 - `conn.countAnnotations()`
 - `conn.listOrphanedAnnotations()`
 - `AnnotationWrapper._getQueryString()`

will-moore:
Currently 1 test failure at https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/60/

```
def testGetAnnotations(self, gatewaywrapper, author_testimg_tiny):
...
# Check links to Tag on the Comment
        for obj_type in ["CommentAnnotation", "Annotation"]:
>           annLinks = list(gatewaywrapper.gateway.getAnnotationLinks(
                obj_type, parent_ids=[ann.getId()]))

        wrapper = KNOWN_WRAPPERS.get(parent_type.lower(), None)
>       link_name = wrapper.ann_link_name()
                    ^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'function' object has no attribute 'ann_link_name'

../../../../.venv3/lib64/python3.11/site-packages/omero/gateway/__init__.py:3556: AttributeError
```

pwalczysko:
A new failing test - freshly from today is https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/62/testReport/OmeroPy.test.integration.gatewaytest.test_get_objects/TestGetObject/testGetObjectsAnnotation/ 


```
cls = <class 'omero.gateway.AnnotationWrapper'>
opts = {'parent_type': 'Dataset', 'parent_ids': [572]}

    @classmethod
    def _getQueryString(cls, opts=None):
        """
        Used for building queries in generic methods such as
        getObjects("Annotation")
        Returns a tuple of (query, clauses, params).
    
        :param opts:        Dictionary of optional parameters.
                            parent_type: (optional) "Project", "Dataset", "Image" etc
                            parent_ids: (optional) list of IDs for the parent type
                            ns: (optional) namespace string to filter by
        :return:            Tuple of string, list, ParametersI
        """
    
        query, clauses, params = super(
            AnnotationWrapper, cls)._getQueryString(opts)
        if opts is None:
            opts = {}
    
        fetch_file = ""
        if cls.OMERO_CLASS in ("Annotation", "FileAnnotation"):
            fetch_file = "left outer join fetch obj.file as file"
    
        query = (f"select obj from {cls.OMERO_CLASS} obj "
                 f"{fetch_file} "
                 "join fetch obj.details.owner as owner "
                 "join fetch obj.details.creationEvent")
    
        # We want to make parent_type case-insensitive...
        if 'parent_type' in opts:
>           ann_link_name = ann_link_name(opts['parent_type'])
                            ^^^^^^^^^^^^^
E           UnboundLocalError: cannot access local variable 'ann_link_name' where it is not associated with a value
```

will-moore:
@sbesson ready for another review, thanks.
I updated the `%sAnnotationLink` logic everywhere that uses it in the BlitzGateway.
So you can now do e.g:

```
        tag = omero.gateway.TagAnnotationWrapper(conn)
        tag.setValue("Test")
        group = conn.getGroupFromContext()
        group.linkAnnotation(tag)

        # also link annotations
        tag.linkAnnotation(anotherAnnotation)

        # listAnnotations works too
        group.listAnnotations()
```
and the `ExperimenterGroupAnnotationLink` and `AnnotationAnnotationLink` objects will be created.

Tests for both of these are in PR at https://github.com/ome/openmicroscopy/pull/6458

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#54](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/54/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/54/consoleText) for more details.
Possible conflicts:
  - PR #459 Tom-TBT 'Wrapper improvements for ROIs'
    - src/omero/gateway/__init__.py
  - Upstream changes
    - src/omero/gateway/__init__.py

--conflicts

will-moore:
@sbesson Do we have any more omero-py PR reviews in the calendar (I'm not seeing any)? Or maybe some other time we can discuss/review this PR? Thx

sbesson:
> Do we have any more omero-py PR reviews in the calendar (I'm not seeing any)? 

That's correct, it looks like the monthly OMERO PR review meeting expired. I propose we discuss this at the next Tuesday meeting. In particular, with the upcoming work towards OMERO.server 5.7.0, we might want to restart them and agree about the frequence and time.

> Or maybe some other time we can discuss/review this PR?

I wasn't sure about the state of this work based on the commits pushed after https://github.com/ome/omero-py/pull/489#issuecomment-5409555903 here and in the companion PR as well as the merge conflicts raised in https://github.com/ome/omero-py/pull/489#event-30263691335. If you are happy with its current state and consider it ready for re-evaluation, I will schedule another round of review probably next week.

will-moore:
Those last commits were fixing additional tests I added at https://github.com/ome/openmicroscopy/pull/6458/changes/1adbba843cc151b98357bf62d98405d15f3f7f65
and also to avoid merge conflicts with https://github.com/ome/omero-py/pull/459 (and because I'd missed `getAnnotationLinkTableName()` which was modified in that PR)

src/omero/gateway/__init__.py
```diff
@@ -904,15 +912,18 @@ def _loadAnnotationLinks(self):
         # Need to set group context. If '-1' then canDelete() etc on
         # annotations will be False
         ctx = self._conn.SERVICE_OPTS.copy()
-        ctx.setOmeroGroup(self.details.group.id.val)
+        if isinstance(self, ExperimenterGroupWrapper):
```
sbesson, line 915: Another wrapper that will need special handling is `ExperimenterWrapper`

```python
>>> e= conn.getObject("Experimenter", 3)
>>> for i in e.listAnnotations():
...     print(i)
... 
  File "<python-input-44>", line 1, in <module>
    for i in e.listAnnotations():
             ~~~~~~~~~~~~~~~~~^^
  File "/Users/sbesson/Documents/GitHub/omero-py/venv/lib/python3.14/site-packages/omero/gateway/__init__.py", line 1022, in listAnnotations
    for ann in self._getAnnotationLinks(ns):
               ~~~~~~~~~~~~~~~~~~~~~~~~^^^^
  File "/Users/sbesson/Documents/GitHub/omero-py/venv/lib/python3.14/site-packages/omero/gateway/__init__.py", line 943, in _getAnnotationLinks
    self._loadAnnotationLinks()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/sbesson/Documents/GitHub/omero-py/venv/lib/python3.14/site-packages/omero/gateway/__init__.py", line 918, in _loadAnnotationLinks
    ctx.setOmeroGroup(self.details.group.id.val)
                      ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'id'
```

src/omero/gateway/__init__.py
```diff
@@ -1039,7 +1054,10 @@ def _linkObject(self, obj, lnkobjtype):
         :type obj:      :class:`BlitzObjectWrapper`
         """
         ctx = self._conn.SERVICE_OPTS.copy()
-        ctx.setOmeroGroup(self.details.group.id.val)
+        if isinstance(self, ExperimenterGroupWrapper):
```
sbesson, line 1059: Same as above

src/omero/gateway/__init__.py
```diff
@@ -10849,6 +10755,10 @@ class _LightSourceWrapper (BlitzObjectWrapper):
               '#type;lightSourceType',
               'version')
 
+    @classmethod
+    def ann_link_name(klass):
+        return "LightSourceAnnotationLink"
```
sbesson, line 10767: This could benefit from a comment similar to `AnnotationWrapper` explaining why we are overriding this

src/omero/gateway/__init__.py
```diff
@@ -11139,6 +11049,15 @@ def refreshWrappers():
...
                            "mapannotation": MapAnnotationWrapper,
+                           "xmlannotation": XmlAnnotationWrapper,
+                           "channel": ChannelWrapper,
+                           "detector": DetectorWrapper,
+                           "dichroic": DichroicWrapper,
+                           "filter": FilterWrapper,
+                           "instrument": InstrumentWrapper,
+                           "lightpath": LightPathWrapper,
+                           "objective": ObjectiveWrapper,
+                           "planeinfo": PlaneInfoWrapper,
```
sbesson, line 11067: Should the light sources concrete instances (filament, arc, laser, lightemittingdiode) be included in the list of known wrappers ?

src/omero/gateway/__init__.py
```diff
@@ -5164,13 +5182,42 @@ def _getQueryString(cls, opts=None):
         Returns a tuple of (query, clauses, params).
 
         :param opts:        Dictionary of optional parameters.
-                            NB: No options supported for this class.
+                            parent_type: (optional) "Project", "Dataset", "Image" etc
```
sbesson: Is it correct to assume any member of `KNOWN_WRAPPERS.keys()` is a valid `parent_type`?

sbesson (changes requested):
This is looking great. The extended usage of `KNOWN_WRAPPERS`, the introduction of the `ann_link_name` convenience function and the removal of the special casing duplicated in various place is a good step forward in terms of maintenance. Same for the consolidation and the generalization of `_getQueryString`.

A couple of inline comments mostly around the usage of wrapper classes that were not part of the initial scope (experimenter & light sources) but can be annotated as per the model.

In addition to the integration tests, could we also expand the unit tests under `test/unit/gatewaytest` to have some coverage of the new `ann_link_name` API for all supported object types. Given the complexity of the gateway code, this will reduce the risk of future regressions.

src/omero/gateway/__init__.py
will-moore: I could add them, but then I feel I'd need to add more tests etc which can be a pain to even create these objects without linking them to others etc.
This is quite a bit beyond the original scope of this PR and we haven't heard users wanting to do `conn.getObject("LightEmittingDiode")` so I feel this could be done as and when needed.

src/omero/gateway/__init__.py
will-moore: Yes, I think so. Even e.g. "TagAnnotation" would give you "AnnotationAnnotationLink".
There's no rule to say that all `KNOWN_WRAPPERS` should be annotatable, but checking against the list above it seems to be that the current list are all annotatable.

src/omero/gateway/__init__.py
will-moore: Done in [972fee6](https://github.com/ome/omero-py/pull/489/commits/972fee6798b178ba02aef8b9cdc7b5993618dd45)

src/omero/gateway/__init__.py
sbesson: Fair enough. As the keys of `KNOWN_WRAPPERS` specified allowed values in multiple APIs including `getAnnotationLinks` but also `getObjects` and `listOrphans`, I appreciate that adding to this list has implications in terms of testing.

The additions of f5b9b5ec0d7993aa61e3c14e0c13208a6baaa19c are slightly confusing with regards to this objective. Why add `lightpath` and `objective` but not `lightemittingdiode`?  In order to bound the scope of this PR, should we limit it to its original scope, push this commit away and only add the missing `xmlannotation` to the dictionary?

src/omero/gateway/__init__.py
```diff
@@ -1039,7 +1056,18 @@ def _linkObject(self, obj, lnkobjtype):
...
-        ctx.setOmeroGroup(self.details.group.id.val)
+        if isinstance(self, ExperimenterGroupWrapper):
+            ctx.setOmeroGroup(self.id)
+        elif isinstance(self, ExperimenterWrapper):
+            # If user is me, save to default group
+            if self._conn.getUserId() == self.id:
+                group_id = self._conn.getEventContext().groupId
+            else:
+                # save link to a group that the user is in
+                group_id = [g.id.val for g in self.copyGroupExperimenterMap()][0]
```
sbesson: An alternative would be `self.getDefaultGroup(self.id).id.val`

sbesson (commented):
One minor inline question about the implementation of `ExperimenterWrapper.linkAnnotation()`.

The integration tests introduced in https://github.com/ome/openmicroscopy/pull/6458 should cover the APIs modified by this PR for all annotation type but were not been included in the nightly CI builds. Once we get a succesful run with the extended integration tests and unit tests are added to this PR as discussed in https://github.com/ome/omero-py/pull/489#pullrequestreview-5224831044 , we should be in a position to get this merged.

knabar (approved):
Don't have anything to add to @sbesson's comments

src/omero/gateway/__init__.py
will-moore: Done in 09328ffb9

src/omero/gateway/__init__.py
will-moore: Added test in 93eb834c1 to test the case-insensitive handling of object names

sbesson (approved):
See https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/96/testReport/OmeroPy.test.integration.gatewaytest/ and https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-test-integration/96/testReport/OmeroPy.test.integration.gatewaytest.test_get_objects/ for the execution of the new integration tests matching these changes.

PR #498, opened 2026-07-30
Upgrade GitHub actions

sbesson:
- apply a routine set of action upgrades in anticipation of the current deprecations - see https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/ and https://github.blog/changelog/2025-10-28-upcoming-deprecation-of-codeql-action-v3/
- bumps the GitHub runner for the tests to ubuntu-latest
- sets some basic GitHub workflow permissions - see https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

Testing should be limited to checking GitHub actions still runs and no longer warns about the deprecated actions

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -37,12 +37,12 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v3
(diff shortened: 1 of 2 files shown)


knabar (approved):
Changes look good, and I don't see any more warnings in the action run

